PR #56, opened 2015-02-03
LayoutRegistry composes Hash instead of inheriting.

### LINKED ISSUE #55: Inheriting from core class, Hash
Was browsing through the code and saw that the `LayoutRegistry` [inherits](https://github.com/lotus/view/blob/07b287168b029564e40cc53e754be58b8e9aa997/lib/lotus/view/rendering/layout_registry.rb#L24) from `Hash`.

Thoughts on composing a `@registry = {}` instead of using inheritance?

https://github.com/lotus/utils/pull/14

jc00ke:
Closes #55


jodosha:
@jc00ke This looks great, thank you! :+1:

PR #58, opened 2015-02-04
Add Slim to list of supported template engines

jc00ke:


jodosha:
@jc00ke Thanks for this addition :+1:

PR #63, opened 2015-03-10
Move scope.__send__ infront of layout.__send__

tomkadwill:
Related to https://github.com/lotus/view/issues/61

### DIFF
--- a/lib/lotus/view/rendering/layout_scope.rb
+++ b/lib/lotus/view/rendering/layout_scope.rb
@@ -154,10 +154,12 @@ module Lotus
         #   # `article` will be looked up in the view scope first.
         #   # If not found, it will be searched within the layout.
         def method_missing(m, *args, &blk)
-          begin
+          if @scope.respond_to?(m)
             @scope.__send__(m, *args, &blk)
-          rescue
+          elsif @layout.respond_to?(m)
             @layout.__send__(m, *args, &blk)
+          else
+            super
           end
         end
 
--- a/test/fixtures.rb
+++ b/test/fixtures.rb
@@ -290,6 +290,10 @@ class ViewForScopeTest
   def bar
     'y'
   end
+
+  def error_inside
+    unknown_method
+  end
 end
 
 module Store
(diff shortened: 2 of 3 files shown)


joneslee85:
@tomkadwill can you please fix the build?

tomkadwill:
@joneslee85 do you or @jodosha have a view on whether this is the right approach to take? If it is not the right approach then there is not much point fixing the tests.

jodosha:
@tomkadwill If it breaks the build, then it's not the right approach. :stuck_out_tongue_winking_eye: 
The order of that lines matter, that's why there is a test to prevent regressions.

I'd say that `Scope` and `LayoutScope` should follow `Lotus::Utils::Hash#method_missing` approach: 1. check if the wrapped object responds to the message (eg. use `respond_to?`) 2. if it fails, instead of calling `super`, raise `NoMethodError` explicitely, blaming the wrapped object instead of the wrapper (`Scope` or `LayoutScope`).

jodosha:
@tomkadwill I agree, the current proposal is hacky. I'm afraid that it won't be portable on other Ruby VMs and it's subject to change in future MRI implementations.

I have a different idea. Because now both `Scope` and `LayoutScope` implement "respond to" logic, we can make this implementation cleaner:

```ruby
def method_missing(m, *args, &blk)
  if @scope.respond_to?(m)
    @scope.__send__(m, *args, &blk)
  elsif @layout.respond_to?(m)
    @layout.__send__(m, *args, &blk)
  else
    super
  end
end
```

If this passes, it's faster, cleaner and should solve the problem.

tomkadwill:
@jodosha Thanks. I've updated the PR with your suggested fix and it seems to work nicely

jodosha:
@tomkadwill Okay, but I think we can call this PR complete once we will have a test that covers this case. I suspect that we could create a regression in the future if we skip a test.

```ruby
exception = -> { scope.unknown }.must_raise NoNethodError
exception.message.must_equal '...'
```

Can you please do it? Thank you :smile:

tomkadwill:
@jodosha I've added 3 tests. The first 2 ensure that my change hasn't introduced any regression and the 3rd tests the new behavious

jodosha:
@tomkadwill Thank you, I've just merged it! :+1:

PR #65, opened 2015-03-18
Fix disable layout in a view

### LINKED ISSUE #64: Can not set null layout for a view inside lotus app
Example (locked on lotusrb 0.2.1):

```ruby
module Web::Views::Mailer
  class Index
    include Web::View
    include Lotus::Helpers

    layout nil # false also fails
  end
end

Web::Views::Mailer::Index.layout # ...::ApplicationLayout
```

AlfonsoUceda:
Closes #64 

What do you think with this approach to disable a layout in a view?

The problem is if you get the layout of a view without setting a new layout, then use configuration.layout in lotus apps

@jodosha @joneslee85 @gzigzigzeo

### DIFF
--- a/lib/lotus/view/dsl.rb
+++ b/lib/lotus/view/dsl.rb
@@ -212,12 +212,13 @@ module Lotus
       end
 
       # When a value is given, it specifies the layout.
+      # When false is given, Lotus::View::Rendering::NullLayout is returned.
       # Otherwise, it returns the previously specified layout.
       #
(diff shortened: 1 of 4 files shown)


jodosha:
@AlfonsoUceda This looks good for me.
@gzigzigzeo Can you please verify if it solves your problem. If so, we can merge it. :+1:

gzigzigzeo:
@jodosha it solves my problem! thank you! :+1:

PR #69, opened 2015-05-21
Render optional contents with #content_for

jodosha:
If we want to render optional contents such as sidebar links or page specific javascripts, we can use `content_for`
It accepts a key that represents a method that should be available within the rendering context.
That context is made of the locals, and the methods that view and layout respond to.
If the context can't dispatch that method, it returns `nil`.

Given the following layout template.

```erb
<!doctype HTML>
<html>
  <!-- ... -->
  <body>
    <!-- ... -->
    <%= content_for :footer %>
  </body>
</html>
```

We have two views, one responds to `#footer` (`Products::Show`) and the other doesn't (`Products::Index`).
When the first is rendered, `content_for` gives back the returning value of `#footer`.
In the other case, `content_for` returns `nil`.

```ruby
module Products
  class Index
    include Lotus::View
  end

  class Show
    include Lotus::View

    def footer
      "contents for footer"
    end
  end
end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -470,6 +470,46 @@ Articles::Index.render(format: :rss)  # => Will use nothing
 
 As per convention, layout templates are located under `Lotus::View.root` or `ApplicationLayout.root` and use the underscored name (eg. `ApplicationLayout => application.html.erb`).
 
+### Content For
+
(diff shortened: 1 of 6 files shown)


jodosha:
After a discussion with @joneslee85 We decided to opt for `#content`. Implemented by #70

PR #70, opened 2015-05-21
Introduced #content as API to render optional contents

### LINKED ISSUE #68: Introduce #content as API to render optional contents
If we want to render optional contents such as sidebar links or page specific javascripts, we can use `content`
It accepts a key that represents a method that should be available within the rendering context.
That context is made of the locals, and the methods that view and layout respond to.
If the context can't dispatch that method, it returns `nil`.

Given the following layout template.

```erb
<!doctype HTML>
<html>
  <!-- ... -->
  <body>
    <!-- ... -->
    <%= content :footer %>
  </body>
</html>
```

We have two views, one responds to `#footer` (`Products::Show`) and the other doesn't (`Products::Index`).
When the first is rendered, `content` gives back the returning value of `#footer`.
In the other case, `content` returns `nil`.

```ruby
module Products
  class Index
    include Lotus::View
  end

  class Show
    include Lotus::View

    def footer
      "content for footer"
    end
  end
end
```

jodosha:
Closes #68

### DIFF
--- a/README.md
+++ b/README.md
@@ -470,6 +470,46 @@ Articles::Index.render(format: :rss)  # => Will use nothing
 
 As per convention, layout templates are located under `Lotus::View.root` or `ApplicationLayout.root` and use the underscored name (eg. `ApplicationLayout => application.html.erb`).
 
+### Content For
+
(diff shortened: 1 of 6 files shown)


joneslee85:
:+1:

PR #71, opened 2015-06-04
Layout methods fix/235

tomkadwill:
Fix for https://github.com/lotus/lotus/issues/235

With this fix the view is able to call methods that defined on the layout


jodosha:
@tomkadwill I've just merged it, thank you. :+1:

PR #77, opened 2015-09-09
Autoescape layout helpers

jodosha:
@lotus/core-team

### DIFF
--- a/lib/lotus/view/rendering/layout_scope.rb
+++ b/lib/lotus/view/rendering/layout_scope.rb
@@ -1,3 +1,5 @@
+require 'lotus/utils/escape'
+
 module Lotus
   module View
     module Rendering
@@ -206,7 +208,7 @@ module Lotus
           elsif layout.respond_to?(m)
             layout.__send__(m, *args, &blk)
           else
-            super
+            ::Lotus::View::Escape.html(super)
           end
         rescue ::NameError
           ::Kernel.raise ::NoMethodError.new("undefined method `#{ m }' for #{ self.inspect }", m)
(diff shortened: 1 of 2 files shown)


test/fixtures.rb
```diff
@@ -339,7 +339,7 @@ class StoreLayout
       include Store::Helpers::AssetTagHelpers
 
       def head
-        %(<meta name="lotusrb-version" content="0.3.1">)
+        Lotus::Utils::Escape::SafeString.new %(<meta name="lotusrb-version" content="0.3.1">)
```
AlfonsoUceda: notice lotus version 0.3.1 in this tag, do you think this tag is useful or secure? I mean, you're telling the world what lotus version you are using.

lib/lotus/view/rendering/layout_scope.rb
```diff
@@ -206,7 +208,7 @@ def method_missing(m, *args, &blk)
           elsif layout.respond_to?(m)
             layout.__send__(m, *args, &blk)
           else
-            super
+            ::Lotus::View::Escape.html(super)
```
AlfonsoUceda: do we need tests for this?

joneslee85:
LGTM :+1:

lib/lotus/view/rendering/layout_scope.rb
jodosha: @AlfonsoUceda Without it, existing tests won't pass.

test/fixtures.rb
jodosha: @AlfonsoUceda I don't understand, this is only an example of a `<meta>` HTML tag. It isn't used by Lotus apps.

The fixture could've been something like this:

```ruby
Lotus::Utils::Escape::SafeString.new %(<meta name="foo" content="bar">)
```

:wink:

PR #80, opened 2015-10-13
Raises an error when rendering a partial from another template.

mrubi:
Raises a Lotus::View::MissingTemplateError when rendering a partial from another template.


joneslee85:
great thanks

mrubi:
To you! Lotus is awesome, thanks!

jodosha:
@mrubi Thank you! :+1:

PR #82, opened 2015-10-31
Subclass errors from a general Lotus module error

### LINKED ISSUE #81: All Lotus view errors inherit directly from StandardError
This means any code rescuing these errors has to either:

* Rescue `StandardError` to rescue all errors in this module, or
* List all errors specifically in advance.

Advice I've been given on the topic by @jodosha is to subclass errors within the Lotus modules so they can be caught at a logical level.

liamdawson:
Closes #81


jodosha:
@liamdawson Hello, and thanks for this contribution. Can you please name it `Error`, instead of `ViewError`? Thanks! :+1:

lib/lotus/view/view_error.rb
```diff
@@ -0,0 +1,16 @@
...
+#
+# This is a base error used to produce Lotus view errors.
+#
+# Allows better rescue options than simply catching `StandardError`.
+#
+# @since 0.4.4
+module Lotus
+  module View
+    class ViewError < ::StandardError
+      def intialize(message)
```
pascalbetz: No need to add that constructor

liamdawson:
@jodosha and @pascalbetz — made your suggested adjustments.

lib/lotus/view/rendering/layout_registry.rb
```diff
@@ -9,7 +10,7 @@ module Rendering
       # This is raised at the runtime when Lotus::Layout cannot find it's template.
       #
       # @since 0.3.0
-      class MissingTemplateLayoutError < ::StandardError
+      class MissingTemplateLayoutError < Lotus::View::Error
```
pascalbetz: Just a thought: is there something like a search path in lotus or what extensions were searched? Then this would be useful information in the exception message.

jodosha:
@liamdawson Thank you, merging this. :+1:

jodosha:
@liamdawson I've cleaned your PR with https://github.com/lotus/view/commit/847b7dad6ebfcbd865fc663261a184bd44e06929, because `MissingTemplateError` and `MissingFormatError` were still inheriting from `::StandardError`.

I also moved all the errors in one file (`lib/lotus/view/errors.rb`) in order to don't require `lotus/view/error` everywhere, and because I wanted to put all the errors on the same namespace depth. For instance, the commit changes `Lotus::View::Rendering::MissingTemplateLayoutError` into `Lotus::View::MissingTemplateLayoutError`.

Can you please follow the same approach for the other frameworks? Thank you very much. :sparkles:

PR #86, opened 2016-01-02
Fix for lotus/lotus issue 185 - Performance issue with rendering a partial view

stevehook:
This is a potential fix for https://github.com/lotus/lotus/issues/185

Partial templates are loaded on start-up and cached in the framework configuration object - Lotus::View.configuration and pulled from there by the PartialFinder at runtime.

I'm not too sure about the cache structure. It's a two-level hash at the moment, so that you can lookup the templates for a given path and then pick the one with the required format.

I have left the old implementation as a fallback for now. Are there any good reasons not to remove it completely (other than the fact that there are a few tests I would need to fix)?


pascalbetz:
@stevehook thanks for the PR. Some remarks/questions:

* did you check the effect of caching partials, performance wise?
* what about reloading modified partials (in development mode)?
* what do you think about loading all templates (layouts, partials, templates) the same way?

joneslee85:
@stevehook great work, would you be able to provide a benchmark too?

stevehook:
Thanks for the comments @pascalbetz.

* I must admit I haven't run any proper performance tests, I should be able to do that tomorrow.
* Reloading I'll need to check on also, I thought that the whole framework got reloaded but I'm not sure about that.
* I think it would be a good idea to load all the templates in the same way, for consistency mainly, as there isn't really a performance issue with non-partials. To be honest I didn't want to attempt too big a change to begin with, but I'd be happy to follow it up when you are happy with the direction it's going in.

pascalbetz:
@stevehook i think it's a good start and its good to start small.

pascalbetz:
@stevehook 

I made a small example to test this, my findings:

i never had a cache hit, the ```@partials``` instance was not the same when loading the partials vs. when looking them up. Hence no performance gain as it always finds/instantiates the template anew. Did not investigate further though.

Can you check again? Or perhaps share your sample app?


About the "unified" mechanism for template loading: i think this is something that @jodosha, @joneslee85, @AlfonsoUceda would need to chime in. What do you guys think?
Currently Templates/Layouts/Partials are preloaded differently. I think that mechanism should be unified so that the 'only' difference is the lookup (partials by ```:partials``` argument on the ```render``` call, templates by action/controller name, layout by layout config).
Not sure how code reloading would work? Would the cache need to deal with that (mtime, if file system) or would shotgun/entr handle this?
Also i think the template loading mechanism should be able to handle templates stored somewhere else than file system (quite useful for CMS applications). Not that it need to be provided by lotus (at this time) but a clean/simple extension point/interface would be great.

lib/lotus/view/rendering/partial_finder.rb
```diff
@@ -42,6 +45,14 @@ def find
         end
 
         protected
+        # @since 0.6.0
+        # @api private
+        def find_cached_template
+          partials = Lotus::View.configuration.partials[[view_template_dir, template_name].join(separator)] ||
```
jodosha: @stevehook Referencing `Lotus::View` here is unsafe. Multiple _"copies"_ of the framework, with different configuration instances are allowed within the same Ruby process.

In Lotus projects we can have `Web::View`, or `Admin::View`, with different set of partials. `Lotus::View` shouldn't have partials in its registry.

---

To solve this problem, you should reference `@view.configuration` (see `#find` in this class).

lib/lotus/view/configuration.rb
```diff
@@ -388,6 +409,7 @@ def reset!
         root             DEFAULT_ROOT
         default_encoding DEFAULT_ENCODING
 
+        @partials   = Hash.new
```
jodosha: @stevehook Not sure if it can work, but what about this?

```ruby
@partials = ::Hash.new {|h,k| h[k] = {} }
```

This should save us one line in `#add_partial`:

```ruby
def add_partial(key, format, partial)
  @partials[key][format] = partial
end
```

See http://ruby-doc.org/core-2.3.0/Hash.html#method-c-new

lib/lotus/view/rendering/partial_finder.rb
```diff
@@ -34,7 +35,9 @@ class PartialFinder < TemplateFinder
         # @since 0.4.3
         # @api private
         def find
-          if path = partial_template_under_view_path
+          if cached_template = find_cached_template
```
jodosha: @stevehook Why do we need this double tier finding? I assume that the load process finds all the partials and put them in the appropriate registry (`Configuration#partials`).

At this point no further modifications are allowed. All the partials should be in that registry. In other words, we want to eager load everything, without falling back to runtime finding of partials.

lib/lotus/view/rendering/partial_finder.rb
```diff
@@ -34,7 +35,9 @@ class PartialFinder < TemplateFinder
...
+            cached_template
+          elsif path = partial_template_under_view_path
```
jodosha: @stevehook As consequence of [my comment above](https://github.com/lotus/view/pull/86/files#r48692258), this `elsif` logic should be removed.

Please ensure that `PartialTemplatesFinder` implements the logic of `partial_template_under_view_path`.

lib/lotus/view/configuration.rb
jodosha: @stevehook Can you please add unit tests for `Configuration`?

See:
  * https://github.com/lotus/view/blob/master/test/view/configuration_test.rb#L128
  * https://github.com/lotus/view/blob/master/test/view/configuration_test.rb#L191
  * https://github.com/lotus/view/blob/master/test/view/configuration_test.rb#L295
  * https://github.com/lotus/view/blob/master/test/view/configuration_test.rb#L331

:+1:

stevehook:
@pascalbetz I'll check my sample app as soon as I get chance. One thing I did have to do was run it with the `--no-code-reloading` option to get any kind of caching working but it sounds like you are hitting a different issue.

lib/lotus/view/rendering/partial_finder.rb
jodosha: @stevehook This is [Inappropriate Intimacy](http://c2.com/cgi/wiki?InappropriateIntimacy) of `Configuration`. This method knows that partials are stored as a Hash and the structure of keys and values.

It could be opportune to expose `Configuration#partial_for`, we can pass all the needed arguments to this new method and let `find_cached_template` to just invoke it.

jodosha:
@stevehook Thank you for this PR. It will give an perf boost to Lotus apps!! :+1: 

We probably need to add an integration test to prove that the whole stack works as expected. We have `Store` and `CardDeck` fixtures (see `test/fixtures.rb`) that can be expanded a bit to prove that:

  * One application (eg. `Store`) is able to render its own partial.
  * Referencing a partial that belongs to another application, will raise an error.

jodosha:
@stevehook The next version of `lotus-view` is `v0.5.0`, that will be released on Jan 12th. However, I see this new logic to be merged later that date.

Can you please replace all the `@since 0.6.0` with `@since x.x.x`? This is our convention for `master` branch. Thank you!

lib/lotus/view/configuration.rb
```diff
@@ -377,9 +379,28 @@ def duplicate
...
         freeze
       end
 
+      # Load partials for each partial template file found under the
+      # given load paths
+      #
+      # @since 0.6.0
+      # @api private
+      def load_partials!
+        Lotus::View::Rendering::PartialTemplatesFinder.new.find_partials(root).each { |key, format, template| add_partial(key, format, template) }
```
pascalbetz: Could you move this to multiple lines with ```do/end```? i think this would be beter for readability.

lib/lotus/view/rendering/partial_finder.rb
stevehook: Thanks for all the helpful comments @jodosha 

This is the part that I found most difficult to understand. I've changed this to use `Lotus::View::Configuration.for(@view)`. Not sure if it is quite right yet.

CC @pascalbetz I think this is the reason that the sample application you created would have not benefitted from the cache.

lib/lotus/view/configuration.rb
stevehook: I've added unit tests for `Configuration`.

Also changed the `@partials` hash as suggested, though it had some knock-on effects in the lookup logic which relied on `@partials[missing_key]` returning a nil. So I'm not sure I really like this change because lookups can mutate the cache.

lib/lotus/view/rendering/partial_finder.rb
stevehook: :+1: 

I've removed the old implementation. To get the tests passing I needed to explicitly reload the configuration.

lib/lotus/view/configuration.rb
stevehook: :+1:

stevehook:
I've updated the PR and addressed some of the comments. I still need to get some performance figures and look at the integration tests among other things.

lib/lotus/view/configuration.rb
```diff
@@ -377,9 +379,29 @@ def duplicate
...
         freeze
       end
 
+      # Load partials for each partial template file found under the
+      # given load paths
+      #
+      # @since x.x.x
+      # @api private
+      def load_partials!
+        Lotus::View::Rendering::PartialTemplatesFinder.new.find_partials(root).each do |key, format, template|
```
joneslee85: what do you think about module or class method instead?

lib/lotus/view/configuration.rb
```diff
@@ -388,6 +410,7 @@ def reset!
         root             DEFAULT_ROOT
         default_encoding DEFAULT_ENCODING
 
+        @partials   = ::Hash.new { |h, k| h[k] = Hash.new }
```
joneslee85: do we need to be explicit with core Hash class here?

pascalbetz:
@stevehook i'll try and unify template loading on top of your changes.

lib/lotus/view/configuration.rb
pascalbetz: while i like the simplified syntax on the lookup side, i think that @stevehook is right when he mentions "lookup mutating the state" as a a downside.

pascalbetz:
@stevehook can you run a benchmark with/without partials, with/without your changes so we can be sure that the change has the desired effect?

stevehook:
I've done some rough benchmarks, using the technique that @smt116 used to diagnose the original problem. The test is hitting a very simple test application (that renders a single partial - https://github.com/stevehook/lotus-test).

Seems to me there is a measurable difference especially with Puma.

branch master with Webbrick:

    $  wrk --connections 4 --duration 5 --threads 4 http://localhost:2300/books
    Running 5s test @ http://localhost:2300/books
      4 threads and 4 connections
      Thread Stats   Avg      Stdev     Max   +/- Stdev
        Latency     9.68ms    3.51ms  25.93ms   69.77%
        Req/Sec   103.76     16.02   141.00     61.50%
      2072 requests in 5.02s, 0.92MB read
    Requests/sec:    413.13
    Transfer/sec:    187.20KB

branch issue_185 with Webbrick:

    $  wrk --connections 4 --duration 5 --threads 4 http://localhost:2300/books
    Running 5s test @ http://localhost:2300/books
      4 threads and 4 connections
      Thread Stats   Avg      Stdev     Max   +/- Stdev
        Latency     6.28ms    2.65ms  31.21ms   69.31%
        Req/Sec   160.58     19.74   232.00     73.50%
      3206 requests in 5.01s, 1.42MB read
    Requests/sec:    639.34
    Transfer/sec:    289.77KB

branch master with Puma:

    $ wrk --connections 4 --duration 5 --threads 4 http://localhost:2300/books
    Running 5s test @ http://localhost:2300/books
      4 threads and 4 connections
      Thread Stats   Avg      Stdev     Max   +/- Stdev
        Latency     6.50ms    2.06ms  19.05ms   72.06%
        Req/Sec   154.54     16.84   191.00     78.00%
      3083 requests in 5.01s, 1.10MB read
    Requests/sec:    615.32
    Transfer/sec:    224.31KB

branch issue_185 with Puma:

    $ wrk --connections 4 --duration 5 --threads 4 http://localhost:2300/books
    Running 5s test @ http://localhost:2300/books
      4 threads and 4 connections
      Thread Stats   Avg      Stdev     Max   +/- Stdev
        Latency     3.88ms    1.30ms  19.90ms   76.02%
        Req/Sec   259.65     25.51   320.00     67.50%
      5177 requests in 5.01s, 1.84MB read
    Requests/sec:   1033.85
    Transfer/sec:    376.73KB

lib/lotus/view/configuration.rb
stevehook: :+1:

`PartialTemplatesFinder` doesn't have any state so makes sense to me.

lib/lotus/view/configuration.rb
stevehook: Removed redundant `::`

joneslee85:
nice, I love what I am seeing :+1:

jodosha:
@stevehook :100: Thank you very much! I just need some more time to review the entire PR. I want to merge this after we'll release v0.6.0 (next Tue).

stevehook:
Thanks @jodosha - I didn't get change to add those integration tests yet. I can try to do that over the next few days.

jodosha:
@stevehook Please do :smile:

stevehook:
@jodosha I was having some trouble with those integration tests.

My first attempt was to extend the existing fixtures as you suggested.  https://github.com/stevehook/view/compare/issue_185...stevehook:issue_185_integration_tests

I extended the fixtures for `CardDeck` and `Store` apps so that each has a view with a partial in their own app, added tests for those together with a test to verify that an exception is raised when attempting to include a partial from the other app. These work fine in isolation but not when running the full test suite because there are other tests that unload these apps. A simple reload doesn't restore the original configuration; the `root` attribute is reset in the `unload!` but not restored in `load!`.

I needed to add a layout template file to render `CardDeck` views - `test/fixtures/templates/card_deck/app/templates/application.html.erb`. This change broke a couple of other tests that render the `Articles::Index` view I think because this new layout file was getting used instead of `test/fixtures/templates/application.html.erb`.

So to avoid these side-effects I ended up creating two new small apps in fixtures (App1 and App2) just to test partials. These seem to work but there may be a simpler way to do this that I'm not seeing at the moment.

jodosha:
@stevehook 

> So to avoid these side-effects I ended up creating two new small apps in fixtures (App1 and App2) just to test partials.

You took the right decision then. :+1: 

---

These problems for our builds are due to global state (`CardDeck` or `Lotus::View`), being (un)loaded. We need to fix those, but after Lotus 1.0 will be out.

lib/lotus/view/rendering/partial_finder.rb
```diff
@@ -34,14 +35,24 @@ class PartialFinder < TemplateFinder
...
-            super
-          end
+          find_cached_template
         end
 
         protected
+
+        # @since x.x.x
+        # @api private
+        def find_cached_template
```
jodosha: @stevehook This abstraction is not needed because `#find` only invokes it and don't do anything else. So `#find_cached_template` can become `#find` instead.

lib/lotus/view/rendering/partial_finder.rb
```diff
@@ -34,14 +35,24 @@ class PartialFinder < TemplateFinder
...
+          conf = Lotus::View::Configuration.for(@view)
```
jodosha: @stevehook `@view` already has its own configuration associated: `@view.configuration`. Doing a lookup with `Configuration.for` is more expensive.

lib/lotus/view/rendering/partial_templates_finder.rb
```diff
@@ -0,0 +1,49 @@
...
+        # @return [Array] array of arrays containing partial template name,
+        # format and a newly created View::Template
+        #
+        # @since x.x.x
+        def self.find_partials(path)
+          _find_partials(path).map do |template|
+            path_name = Pathname(template)
+            partial_path, partial_base_name = Pathname(template).relative_path_from(path).split
+            partial_base_parts = partial_base_name.to_s.split('.')
+            ["#{partial_path}#{::File::SEPARATOR}#{partial_base_parts[0]}", partial_base_parts[1], View::Template.new(template)]
```
jodosha: @stevehook We have lost encoding informations here. It should be `View::Template.new(template, default_encoding)`, where `default_enconding` is taken from `Configuration#default_enconding`.

jodosha: @stevehook You should probably let `Configuration` to invoke this class like the following snippet, so you can pass easily access the `Configuration#default_encoding` information.

```ruby
Lotus::View::Rendering::PartialTemplatesFinder.new(self, root).find
```

In the code above `self` is an instance of `Configuration`. This refers to the new implemented method `Configuration#load_partials!`.

lib/lotus/view/configuration.rb
```diff
@@ -377,9 +379,39 @@ def duplicate
...
         freeze
       end
 
+      # Load partials for each partial template file found under the
+      # given load paths
+      #
+      # @since x.x.x
+      # @api private
+      def load_partials!
+        Lotus::View::Rendering::PartialTemplatesFinder.find_partials(root).each do |key, format, template|
```
jodosha: @stevehook This set of `key`, `format`, `template` smells like a [Data Clump](http://martinfowler.com/bliki/DataClump.html). Is it reasonable to yield just an object named `Partial`.

This would also simplify the signature of `#add_partial(partial)`.

lib/lotus/view/rendering/partial_finder.rb
stevehook: @jodosha using `@view.configuration` doesn't work for me. This area is still confusing me but the partials are being attached to the configuration object on the application namespace and not being copied down to the configurations for the individual views. I see why this is expensive but not sure how to resolve it in a more efficient way.

I think I've fixed all the other issues you've raised.

jodosha: @stevehook I see and sorry for suggesting a wrong solution. If this works and not the other one, we should use your solution. Thank you :+1: 

One last thing before to merge. Because the code has changed since the last benchmark, can we run it again and compare? Thanks :100: :sparkling_heart:

stevehook:
I've rerun the simple performance benchmark using latest master and issue_185 branches, results below.

It looks like the results are pretty consistent with the original measurements I took before the latest changes, neither master or this branch have got any better or worse.

branch master with Webrick:

    wrk --connections 4 --duration 5 --threads 4 http://localhost:2300/books
    Running 5s test @ http://localhost:2300/books
      4 threads and 4 connections
      Thread Stats   Avg      Stdev     Max   +/- Stdev
        Latency     9.20ms    3.08ms  21.89ms   70.32%
        Req/Sec   108.93     13.61   141.00     73.00%
      2180 requests in 5.03s, 0.96MB read
    Requests/sec:    433.63
    Transfer/sec:    196.56KB

branch issue_185 with Webrick:

    wrk --connections 4 --duration 5 --threads 4 http://localhost:2300/books
    Running 5s test @ http://localhost:2300/books
      4 threads and 4 connections
      Thread Stats   Avg      Stdev     Max   +/- Stdev
        Latency     6.25ms    2.54ms  19.51ms   66.78%
        Req/Sec   161.29     17.80   202.00     59.50%
      3225 requests in 5.02s, 1.43MB read
    Requests/sec:    642.17
    Transfer/sec:    291.12KB

branch master with Puma:

    wrk --connections 4 --duration 5 --threads 4 http://localhost:2300/books
    Running 5s test @ http://localhost:2300/books
      4 threads and 4 connections
      Thread Stats   Avg      Stdev     Max   +/- Stdev
        Latency     6.67ms    1.78ms  16.38ms   71.65%
        Req/Sec   150.43     13.40   200.00     80.50%
      3003 requests in 5.01s, 1.07MB read
    Requests/sec:    599.08
    Transfer/sec:    218.27KB

branch issue_185 with Puma:

    wrk --connections 4 --duration 5 --threads 4 http://localhost:2300/books
    Running 5s test @ http://localhost:2300/books
      4 threads and 4 connections
      Thread Stats   Avg      Stdev     Max   +/- Stdev
        Latency     3.94ms    1.15ms  10.36ms   73.10%
        Req/Sec   254.18     17.33   373.00     79.00%
      5068 requests in 5.01s, 1.80MB read
    Requests/sec:   1011.78
    Transfer/sec:    368.67KB

lib/lotus/view/rendering/partial_finder.rb
stevehook: @jodosha no problem, thank you for all the helpful comments! I've rerun the benchmarks, results in comment below.

jodosha:
@stevehook I've finally merged this. Thank you very much! :+1:

stevehook:
thank you @jodosha and love the new project name by the way!

PR #88, opened 2016-01-08
Guarantee default encoding to exist

jodosha:
Ditto.

---

Default encoding is now `Encoding::UTF_8`, instead of `"utf-8"`.
Developers can set the value with a string:

```ruby
Lotus::View.configure do
  default_encoding "ISO-8859-1"
end
```

or with a constant:

```ruby
Lotus::View.configure do
  default_encoding Encoding::ISO_8859_1
end
```

Invalid names will cause an `ArgumentError`

```ruby
Lotus::View.configure do
  default_encoding 'foo'
end

# => ArgumentError: unknown encoding name - foo
```

This _"fail fast"_ policy let developers to detect unsupported encodings at the startup time (in development or production), rather than later on when the application is up and running.

---

Thanks to @deepj for letting me to reflect on the poor handling that we had until now.

Ref: https://github.com/lotus/view/commit/2fe93731e5f84cd34de0d68e320b2e35db671a4a#commitcomment-15324075

/cc @lotus/core-team

### DIFF
--- a/lib/lotus/view/configuration.rb
+++ b/lib/lotus/view/configuration.rb
@@ -33,7 +33,7 @@ module Lotus
       #
       # @since 0.5.0
       # @api private
-      DEFAULT_ENCODING = 'utf-8'.freeze
+      DEFAULT_ENCODING = Encoding::UTF_8
 
       attr_reader :load_paths
       attr_reader :views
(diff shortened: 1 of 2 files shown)


AlfonsoUceda:
perfect :+1: thanks @deepj ;)

deepj:
:+1: :grinning:

PR #90, opened 2016-01-08
Comment in registry.rb#48 notes that there should only be one registr…

pascalbetz:
https://github.com/lotus/view/issues/89

### DIFF
--- a/test/fixtures.rb
+++ b/test/fixtures.rb
@@ -4,6 +4,17 @@ class HelloWorldView
   include Lotus::View
 end
 
+class SuperclassView
+  include Lotus::View
+end
+
+class SubclassView < SuperclassView
+end
+
+class SubclassViewXml < SuperclassView
+  format :xml
+end
+
 class DisabledLayoutView
   include Lotus::View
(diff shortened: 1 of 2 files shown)


joneslee85:
please fix the build

pascalbetz:
@joneslee85 this is a test to show the problem. I would need some input on how to fix it (make sure there is only one registry per view hierarchy or lifting this restriction and having one registry per view). Probably best to discuss in the issue.

joneslee85:
@pascalbetz sure, bring it to the chat room please

jodosha:
@pascalbetz Hey, now I understand the problem. As I mentioned before, if there is a way to change the internals of this framework, but without changing the existing behaviors, I'm fine with all the solutions.

Do you still want to take care of this?

pascalbetz:
@jodosha i can look into, yes. 
After our previous discussions i'm not sure how much of an issue it is because it was never meant to subclass and keep the format.

jodosha:
@pascalbetz Closing this because not high priority.

PR #93, opened 2016-01-29
Disable autoescape for Slim as already implemented by us

jodosha:
Closes https://github.com/hanami/assets/issues/36


davydovanton:
amazing! :+1:

PR #94, opened 2016-02-20
Fix SimpleCov deprecation message

deepj:
This avoids the following deprecation message:

```
test/test_helper.rb:8:in `<top (required)>': [DEPRECATION] ::[] is
deprecated. Use ::new instead.
```


deepj:
Weird problem with coverage and JRuby on Travis. This commit doesn't change anything in Travis and JRuby settings.

PR #96, opened 2016-03-09
Try local

jodosha:
## Feature

Introducing `#local` as a public method that is available in views, layouts and templates.
In case we want to safely navigate methods or locals, we can use this new method.

Here's some examples.

### Optional Content

Let's say we want to render optional contents for footer, for instance per page javascripts.

```erb
# apps/web/templates/application.html.erb
<!doctype HTML>
<html>
  <!-- ... -->
  <body>
    <!-- ... -->
    <%= local :footer %>
  </body>
</html>
```

We have one view that **DOES NOT** respond to `#footer`

```ruby
# apps/web/views/products/index.rb
module Web::Views::Products
  class Index
    include Web::View
  end
end
```

And another view that **DOES** respond to `#footer`.

```ruby
# apps/web/views/products/show.rb
module Web::Views::Products
  class Show
    include Web::View

    def footer
      javascript 'product'
    end
  end
end
```

When the first view is rendered, `local :footer` is completely ignored and that interpolation returns a blank string. When the second one is rendered, because it respond to `#footer`, that method is invoked and the returning value is used for the interpolation (a `<script>` tag, in the example).

### Optional Locals

We have a view that renders a form.
We may want to show validation errors, only if we have a `result` local that carries on the errors.

```ruby
# apps/web/views/products/new.rb
module Web::Views::Products
  class New
    include Web::View

    def error
      local(:result).error
    end
  end
end
```

```erb
<% unless error.nil? %>
  <%= error.message %>
<% end %>
```

### Safe Locals Navigation

We have announcements to show in our UI.
They may or may not be loaded from the database ad passed to the view.

```erb
<% if local(:announcement).display? %>
  <%= announcement.message %>
<% end %>
```

## Deprecation

We have a similar API that this PR aims to deprecate: `#content`.
The problem with content is that it doesn't allow "Safe Navigation" feature.

For instance

```erb
<% if content(:announcement).display? %>
  <%= announcement.message %>
<% end %>
```

The first line would raise a `NoMethodError`, because `#content` returns `nil` in case the lookup fails. On the other hand, `#local` returns a null object that allows Safe Navigation.

---

/cc @hanami/core-team @hanami/contributors

### DIFF
--- a/lib/hanami/view/rendering.rb
+++ b/lib/hanami/view/rendering.rb
@@ -1,5 +1,6 @@
 require 'hanami/view/rendering/registry'
 require 'hanami/view/rendering/scope'
+require 'hanami/view/rendering/null_local'
 
 module Hanami
   module View
@@ -102,6 +103,20 @@ module Hanami
         #   view.render # => {title: ...}
         def render
           layout.render
+        end
+
+        # Return a local for the corresponding key, otherwise it returns a null object
+        #
+        # @return [Objeect,Hanami::View::Rendering::NullLocal] the returning value
+        #
+        # @since x.x.x
+        #
+        # @example
+        #   <% if local(:plan).overdue? %>
+        #     <h2>Your plan is overdue.</h2>
+        #   <% end %>
+        def local(key)
+          locals.fetch(key) { NullLocal.new(key) }
         end
 
         protected
--- a/lib/hanami/view/rendering.rb
+++ b/lib/hanami/view/rendering.rb
@@ -105,7 +105,8 @@ module Hanami
           layout.render
         end
 
-        # Return a local for the corresponding key, otherwise it returns a null object
+        # It tries to invoke a method for the view or a local for the given key.
+        # If the lookup fails, it returns a null object.
         #
         # @return [Objeect,Hanami::View::Rendering::NullLocal] the returning value
         #
@@ -116,7 +117,11 @@ module Hanami
         #     <h2>Your plan is overdue.</h2>
         #   <% end %>
         def local(key)
-          locals.fetch(key) { NullLocal.new(key) }
+          if respond_to?(key)
+            __send__(key)
+          else
+            locals.fetch(key) { NullLocal.new(key) }
+          end
         end
 
         protected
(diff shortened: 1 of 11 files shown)


vyper:
Cool and very useful! :+1:

I already needed this in my project. (:

lib/hanami/view/rendering.rb
```diff
@@ -104,6 +105,25 @@ def render
...
+        #
+        # @return [Objeect,Hanami::View::Rendering::NullLocal] the returning value
+        #
+        # @since x.x.x
+        #
+        # @example
+        #   <% if local(:plan).overdue? %>
+        #     <h2>Your plan is overdue.</h2>
+        #   <% end %>
+        def local(key)
```
davydovanton: What do you think can we remove or replace duplicate logic to `Rendering#local`?

test/view/rendering/null_local_test.rb
```diff
@@ -0,0 +1,46 @@
...
+  it 'returns false to any?' do
+    @null.any?.must_equal false
+  end
+
+  it 'returns true to empty?' do
+    @null.empty?.must_equal true
+  end
+
+  it 'returns true to nil?' do
+    @null.nil?.must_equal true
```
davydovanton: I think we can use `must_be_nil` matcher here, what do you think?

davydovanton:
amazing feature :+1:

hlegius:
Great job! :ok_hand:

lib/hanami/view/rendering.rb
jodosha: @davydovanton `Rendering#respond_to?` isn't accurate.

If you remove this method, the other one takes precedence but the test fail. It isn't able to check if a local or a method exposed in a layout is available to the entire rendering process, hence the duplication. (sigh).

test/view/rendering/null_local_test.rb
jodosha: @davydovanton `NullLocal` inherits from Ruby's `BasicObject`.

In the past I got odd testing results because Minitest assumes that the current object has a set of features that the descendants of `Object` have.

The difference here is subtle, but intended.

With this syntax, the object under test is `@null`. But we just said that its nature may lead to unexpected testing results.

```ruby
@null.must_be_nil
```

If you look closer, you're asking if `@null` is `nil`, but does it? NO.

We want to test if `@null`, respond to `#nil?` and returns `true`.

```ruby
@null.nil?.must_equal true
```

In this case, the object under test is the returning value of `#nil?`: a boolean, which is safer to test.

Issue #100, opened 2016-05-14
TemplatesFinder has the same issue as the Partials Finder used to have.


nessur:
Nearly identical to the issue fixed here: https://github.com/hanami/view/pull/73

This code is ambiguous: 
`Hanami::View::Rendering::TemplatesFinder.new(Articles::Show).find`
When these files exist:
`apps/web/templates/members/articles/show.html.erb`
`apps/web/templates/articles/show.html.erb`

B/c the `def _find` method just calls:
`Dir.glob( "#{ [root, lookup, template_name].join(separator) }.#{ format }.#{ engines }" )`

My classes look like this:
```
module Web::Views::Articles
  class Index
    include Web::View
    template 'articles/index'
  end
end
```
and
```
module Web::Views::Members::Articles
  class Index
    include Web::View
    template 'members/articles/index'
  end
end
```

This is the Great Understanding I came to when I saw my app work successfully locally, but fail when deployed, I suppose the directories were created in a different order, such that the wrong one was always returned first on the remote server.

jodosha:
Fixed by #101

PR #101, opened 2016-05-14
Fix Nested Resource support for templates

nessur:
If `../templates/members/articles/index.html.erb` and `../templates/articles/index.html.erb` both exist, then the template finder will find the former instead of the latter when grabbing the file for a Articles::Index View class. 

re: issue https://github.com/hanami/view/issues/100

### DIFF
--- a/lib/hanami/view/rendering/templates_finder.rb
+++ b/lib/hanami/view/rendering/templates_finder.rb
@@ -79,7 +79,18 @@ module Hanami
         # @api private
         # @since 0.4.3
         def _find(lookup = search_path)
-          Dir.glob( "#{ [root, lookup, template_name].join(separator) }.#{ format }.#{ engines }" )
+          primary_templates = templates_in_base_path
+          if primary_templates.none?
+            Dir.glob( "#{ [root, lookup, template_name].join(separator) }.#{ format }.#{ engines }" )
+          else
+            primary_templates
+          end
+        end
+
+        # @api private
+        # @since 0.7.0
+        def templates_in_base_path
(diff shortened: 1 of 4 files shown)


lib/hanami/view/rendering/templates_finder.rb
```diff
@@ -79,7 +79,18 @@ def find
         # @api private
         # @since 0.4.3
         def _find(lookup = search_path)
-          Dir.glob( "#{ [root, lookup, template_name].join(separator) }.#{ format }.#{ engines }" )
+          primary_templates = templates_in_base_path
```
davydovanton: why do you create this variable?

davydovanton: oh, I sorry, I found this

davydovanton:
Thanks for contribution 👍 

I want to suggest other implementation. I think that we need to replace path generation logic to separate method. What do you think about this solution?

```ruby
def _find(lookup = search_path)
  base_path = templates_path(root, template_name)

  if base_path.none?
    templates_path(root, lookup, template_name)
  else
    base_path
  end
end

def templates_path(*parts)
  Dir.glob("#{ parts.join(separator) }.#{ format }.#{ engines }")
end
```

zipcar-codeclimate:
Thanks for the suggestion! Great way to remove some code duplication.

nessur:
Thanks for the suggestion! Great way to remove some duplication, makes the code more clear.

jodosha:
@nessur Thank you, I just merged this!

PR #104, opened 2016-06-20
Maintain the original method visibility on Module#prepend

### LINKED ISSUE #102: Hanami::Presenter turns private methods into public ones
Consider the following example:

```ruby
class AwesomePresenter
  include Hanami::Presenter

  private

  def a_method
    puts "You should not pry in the private lives of others"
  end
end

contrived_example = AwesomePresenter.new(Object.new)
contrived_example.public_methods.include? :a_method      #=> true
contrived_example.a_method    #=> "You should not pry in the private lives of others"
```

What I expected is that it would return false and not being able to be called with a receiver other than `self` as `SimpleDelegator` does.

Also, is there a reason that it does not use `SimpleDelegator` or tries to implement `Delegator` at all?

mereghost:
This solves issue #102 where private and protected methods became public
on the prepended modules used for autoescaping.

I'm not that much happy with the code, but it works and isn't that ugly. =)

---

Closes #102

### DIFF
--- a/lib/hanami/view/escape.rb
+++ b/lib/hanami/view/escape.rb
@@ -165,8 +165,9 @@ module Hanami
       # @since 0.4.0
       # @api private
       def method_added(method_name)
+        visibility = :public
         visibility = :private if private_method_defined? method_name
-        visibility ||= :protected if protected_method_defined? method_name
+        visibility = :protected if protected_method_defined? method_name
 
         unless autoescape_methods[method_name]
           prepend Module.new {
--- a/lib/hanami/view/escape.rb
+++ b/lib/hanami/view/escape.rb
@@ -165,10 +165,13 @@ module Hanami
       # @since 0.4.0
       # @api private
       def method_added(method_name)
+        visibility = :private if private_method_defined? method_name
+        visibility ||= :protected if protected_method_defined? method_name
+
         unless autoescape_methods[method_name]
           prepend Module.new {
             module_eval %{
-              def #{ method_name }(*args, &blk); ::Hanami::View::Escape.html super; end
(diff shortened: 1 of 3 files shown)


lib/hanami/view/escape.rb
```diff
@@ -165,10 +165,13 @@ def self.extended(base)
       # @since 0.4.0
       # @api private
       def method_added(method_name)
+        visibility = :private if private_method_defined? method_name
+        visibility ||= :protected if protected_method_defined? method_name
```
davydovanton: Could you explain why you use `||=` here?

mereghost: I'm not really proud of it, but it goes as this: if visibility was set to anything truthy before, skip the assignment. That being said, the first assignment could also be a or_equals `||=` but it ended up even more cryptic. =(

I'm all open to suggestions on a better way to do that (I haven't found one, tho).

cllns: Can you write some tests that leverage this distinction? They will help explain this, ensure there are no regressions in the future, and who knows... might even help you find a simpler solution :D

mereghost: I could change that to an `if .. elsif ... end` which I find particularly ugly or to a `case`.

There isn't a way that a method with the same name is at the same time `private` and `protected` and `public`.

I could make it look like this:

```ruby
visibility = :public
visibility = :private if private_method_defined? method_name
visibility = :protected if protected_method_defined? method_name
```
Does that make it clear enough? (I like this lil' a bit more).

jodosha:
@mereghost Thanks for this fix! I just merged it. It works great.

PR #105, opened 2016-07-22
Ensure to use new registry when loading the framework

jodosha:
Fixes https://github.com/hanami/hanami/issues/619


bruz:
Thanks for this fix @jodosha!

PR #112, opened 2016-11-02
Make second arg to Hanami::View initialize accept keyword arguments

vladra:
Hi,

What do you think about this tiny change to `initialize` method of Hanami::View?

It doesn't change any behavior of `hanami-view` gem, but allows to integrate it with `dry-system` and `dry-auto_inject` gems. Since hanami is cooperating with dry-rb ecosystem, it could be useful I think. At least for me :)

### DIFF
--- a/lib/hanami/view/rendering.rb
+++ b/lib/hanami/view/rendering.rb
@@ -36,7 +36,7 @@ module Hanami
         #
         #   template = Hanami::View::Template.new('index.html.erb')
         #   view     = IndexView.new(template, {article: article})
-        def initialize(template, locals)
+        def initialize(template, **locals)
           @template = template
           @locals   = locals
           @scope    = Scope.new(self, @locals)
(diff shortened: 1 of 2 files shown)


jodosha:
@vladra Hello! Can you please provide a test for it? Thanks!

AlfonsoUceda:
ping @vladra

vladra:
I've seen this. Will try to add some test by the end of this week.

vladra:
Was going to add tests, but isn't it already tested?

Every time `render` is called in `rendering_test.rb`, like in this example:
https://github.com/hanami/view/blob/master/test/rendering_test.rb#L109

it is initializing view with given arguments:
https://github.com/hanami/view/blob/master/lib/hanami/view/rendering/registry.rb#L93

Anyway, I've added small test to test initialize method. Let me know if you see this in a different way

joneslee85:
👍 LGTM

AlfonsoUceda:
LGTM too

Issue #120, opened 2017-03-20
Migrate from Minitest to RSpec


jodosha:
A few rules to follow:

  * If you want to work on this, please leave a comment here. We want to avoid many people to work on the same task.

  * We expect this migration to happen within a week from now.

  * Do not change the tests, only migrate from Minitest to RSpec syntax. If not sure, please ask here.

  * Do not remove the old tests, only add the new ones.

  * Make CI build to use RSpec tests.

  * **Target `develop` branch, instead of `master`.**

  * All the tests artefacts like fixtures, goes under `spec/support`.

  * Try to reflect the structure under `lib/`. Eg. `lib/hanami/view.rb` should be tested under `spec/hanami/view_spec.rb`, while `test/rendering_test.rb` should go under `spec/integration/rendering_spec.rb`.

If you want to take inspiration, please have a look at: https://github.com/hanami/mailer/pull/67

jodosha:
@nerdinand is in charge for this.

sahithi-ihtihas:
@jodosha   I am beginner and would like to take up the issue. Can you please guide me in doing that ?

jodosha:
@sahithi-ihtihas I'm sorry but @nerdinand is already working on this. Please see #121.

Would you like to take another framework? See http://bit.ly/hanami-help-wanted

jodosha:
Implemented by #121

PR #121, opened 2017-03-20
Replace minitest with rspec

nerdinand:
Replace all minitest tests with RSpec. See https://github.com/hanami/view/issues/120


AlfonsoUceda:
Hi nerdinand, thanks for the PR but what is the purpose of this PR, can you explain a little? thanks

nerdinand:
Hi @AlfonsoUceda 

I was talking to @jodosha on Gitter and he told me that this is a way I could get started with doing some work on Hanami. His plan is to move all tests to rspec eventually.

jodosha:
@nerdinand Is there anything we can help with? /cc @hanami/community

nerdinand:
@jodosha yes

Specs are failing on travis, but not not locally at the moment. See: https://travis-ci.org/hanami/view/jobs/213129478 To me it looks like the code is trying to render a partial from the test suite of slim (installed in vendor/bundle) instead of rendering a fixture partial. Of course then the test fails. I guess we have to restrict the search path for partials to the fixture directory? Not sure how to do that though.

There are a couple of failing specs which I have put on pending to make the build pass on travis. Those will have to be looked at. I have tried to explain in the pending message what I think the problem is. I'm looking for input on whether to rewrite those specs entirely or ideas on how to fix them as they are.

nerdinand:
I did some further digging and I could reproduce the failure on travis by creating the file in question. To reproduce, just create a file `/bundle/ruby/2.3.0/gems/slim-3.0.7/test/rails/app/views/layouts/application.html.slim` containing:

```
doctype html

html
  head
    title Dummy
    = csrf_meta_tag

  body
    = yield :page_heading
    .content= yield
```

Because I don't know in detail how the code works yet, I couldn't yet figure out why the layout would get resolved in this directory. It smells a bit like a bug to me, but maybe I'm mistaken...

nerdinand:
It seems to have to do with the way the `root` in `Hanami::View::Configuration` is handled. When I change 
```
DEFAULT_ROOT = '.'.freeze
```

to 
```
DEFAULT_ROOT = '/Users/ferdi/vendor/hanami/view/spec/support/fixtures/templates'.freeze
```

then the test passes. Also, some of the other pending specs don't fail after this change. Any input in helping debug this is welcome!

jodosha:
@nerdinand Please make them all to pass, even if that means to change the `DEFAULT_ROOT`. I suspect this may be triggered by the global state that is changed by some spec during the setup.

nerdinand:
If you look past the hack at https://github.com/hanami/view/pull/121/commits/1e03849d736a97e2fd9a95e599d5ecdf36f9a56b#diff-57884df67a74b681768b6f28bac5fd57R31, there are now 2 kinds of pending specs left to fix:

### `./spec/hanami/view_spec.rb:391`
For some reason, `==` seems to work differently in Minitest and RSpec.

In this example:
```
  3) Hanami::View unload .configuration a view inherits the parent
     # Does == mean something different in rspec and minitest?
     Failure/Error: expect(child).to eq(parent)

       expected: #<Hanami::View::Configuration:0x007fb23029cc98 @namespace=Object, @root=#<Pathname:/Users/ferdi/vendo...thname:/Users/ferdi/vendor/hanami/view/spec/support/fixtures/templates>]>, @layout=nil, @modules=[]>
            got: #<Hanami::View::Configuration:0x007fb2302968e8 @namespace=Object, @root=#<Pathname:/Users/ferdi/vendo...thname:/Users/ferdi/vendor/hanami/view/spec/support/fixtures/templates>]>, @layout=nil, @modules=[]>

       (compared using ==)

       Diff:
       @@ -1,4 +1,4 @@
       -#<Hanami::View::Configuration:0x007fb23029cc98
       +#<Hanami::View::Configuration:0x007fb2302968e8
         @default_encoding=#<Encoding:UTF-8>,
         @layout=nil,
         @layouts=#<Set: {}>,
```

Maybe not the objects themselves but their attributes should get compared? E.g. `load_paths`, `views`, etc.?

### `./spec/hanami/view/configuration_spec.rb:71`
`Hanami::Utils::LoadPaths` does not implement `#include?` which these specs are trying to call. I'm unsure how this worked in Minitest. There are two possible fixes:

* Implement `#include?` for `Hanami::Utils::LoadPaths` (delegate to the `paths` private member)
* Change the specs to call `send(:paths).include?`, e.g. `expect(@configuration.load_paths.send(:paths)).to include '..'`

What do you prefer?

morrme:
@nerdinand have you tried something like expect(child).to be(parent) ?

jodosha:
@morrme would you like to help with the process? /cc @nerdinand 

Thank you all!

morrme:
@jodosha  @nerdinand  yes, how can i help?

jodosha:
@nerdinand @morrme Would you folks like to pair? Or do you want to check the errors and try to apply a patch?

nerdinand:
@jodosha @morrme I think pairing would make sense. Or you could try to investigate yourself, maybe you find out something more... But in the end I think someone needs to make an executive decision about how to fix the specs...

morrme:
@nerdinand did you try my suggestion with be? @jodosha suggested it to me as a way to mend a similar issue and it worked.

jodosha:
@nerdinand @morrme Again, if you can pair it would be great.

If not possible, please try to understand what's the issue with the RSpec migration. I checked the code locally, it looks like the root isn't reset properly by the tests. Could it be that the _shared context_ isn't work properly? Is it worth to add an assertion inside that `before` block, so we can make it to fail early and catch logical errors in that reset logic.

Thank you very much for your patience, this is more challenging that we thought, but still crucial for the progress of new Hanami features.

morrme:
@jodosha I like your idea.

morrme:
@jodosha @nerdinand for the == issue, one thing i tried was to compare attributes of the object, like in the above example, comparing the value for `@root` instead of comparing the objects.  in the end, for my tests,` be` worked fine, but if it doesn't work for you, that may help.

nerdinand:
@morrme `expect(child).to be(parent)` doesn't work here. AFAIK `be` checks that the object is actually the same instance, which we don't want in this case. We want that it has the same attributes, not to `be` the same object:

```ruby
      it 'a view inherits the parent' do
        pending('Does == mean something different in rspec and minitest?')
        parent = ConfigurationView.configuration
        child  = ConfigurationChildView.configuration

        expect(child).to eq(parent)
        expect(child).to_not be(parent)
      end
```

nerdinand:
@morrme @jodosha  I fixed `view_spec.rb` with the suggested approach. Please review https://github.com/hanami/view/pull/121/commits/5dd1a1129250b5a5d769e0ebeeab32af67981bf5 as this might alter the test semantics slightly.

spec/hanami/view_spec.rb
```diff
@@ -381,19 +381,17 @@ class ConfigurationLayout
       end
 
       it 'a view inherits the configuration from the framework' do
-        pending('Does == mean something different in rspec and minitest?')
         expected = Hanami::View.configuration
         actual   = ConfigurationView.configuration
 
-        expect(actual).to eq(expected)
+        expect(actual.root).to eq(expected.root)
```
jodosha: @nerdinand This doesn't consider all the other settings in the configuration.

Here we're exercising `Configuration#==`, so it's worth to make this assertion explicit, instead of relying on the test framework implicit comparison. This should work:

```ruby
expect(actual == expected).to be(true)
```

nerdinand: Sadly, that does not work. I think `expect(actual == expected).to be(true)` is equivalent to `expect(actual).to eq(expected)`.

Also makes sense while debugging:

```
[384, 393] in /Users/ferdi/vendor/hanami/view/spec/hanami/view_spec.rb
   384:         expected = Hanami::View.configuration
   385:         actual   = ConfigurationView.configuration
   386:
   387:         require 'byebug'; byebug
   388:
=> 389:         expect(actual == expected).to be(true)
   390:       end
   391:
   392:       it 'a view inherits the parent' do
   393:         parent = ConfigurationView.configuration
(byebug) expected
#<Hanami::View::Configuration:0x007fc1e5c86690 @namespace=Object, @root=#<Pathname:/Users/ferdi/vendor/hanami/view/spec/support/fixtures/templates>, @default_encoding=#<Encoding:UTF-8>, @partials={}, @views=#<Set: {ConfigurationView}>, @layouts=#<Set: {ConfigurationLayout}>, @load_paths=#<Hanami::Utils::LoadPaths:0x007fc1e5c859c0 @paths=[#<Pathname:/Users/ferdi/vendor/hanami/view/spec/support/fixtures/templates>]>, @layout=nil, @modules=[]>
(byebug) actual
#<Hanami::View::Configuration:0x007fc1e5c85038 @namespace=Object, @root=#<Pathname:/Users/ferdi/vendor/hanami/view/spec/support/fixtures/templates>, @default_encoding=#<Encoding:UTF-8>, @partials={}, @views=#<Set: {}>, @layouts=#<Set: {}>, @load_paths=#<Hanami::Utils::LoadPaths:0x007fc1e5c84980 @paths=[#<Pathname:/Users/ferdi/vendor/hanami/view/spec/support/fixtures/templates>]>, @layout=nil, @modules=[]>
```

There are differences in the attributes of the configurations. However, when running the same test from before migration to rspec:

```
[37, 46] in /Users/ferdi/vendor/hanami/view/test/view_test.rb
   37:       expected = Hanami::View.configuration
   38:       actual   = ConfigurationView.configuration
   39:
   40:       require 'byebug'; byebug
   41:
=> 42:       actual.must_equal(expected)
   43:     end
   44:
   45:     it 'a view inheriths the parent' do
   46:       parent = ConfigurationView.configuration
(byebug) expected
#<Hanami::View::Configuration:0x007ff84925d8e0 @namespace=Object, @root=#<Pathname:/Users/ferdi/vendor/hanami/view/spec/support/fixtures/templates>, @default_encoding=#<Encoding:UTF-8>, @partials={}, @views=#<Set: {ConfigurationView}>, @layouts=#<Set: {ConfigurationLayout}>, @load_paths=#<Hanami::Utils::LoadPaths:0x007ff84925c760 @paths=[#<Pathname:/Users/ferdi/vendor/hanami/view/spec/support/fixtures/templates>]>, @layout=nil, @modules=[]>
(byebug) actual
#<Hanami::View::Configuration:0x007ff849257d00 @namespace=Object, @root=#<Pathname:/Users/ferdi/vendor/hanami/view/spec/support/fixtures/templates>, @default_encoding=#<Encoding:UTF-8>, @partials={}, @views=#<Set: {}>, @layouts=#<Set: {}>, @load_paths=#<Hanami::Utils::LoadPaths:0x007ff8492573f0 @paths=[#<Pathname:/Users/ferdi/vendor/hanami/view/spec/support/fixtures/templates>]>, @layout=nil, @modules=[]>
```

There seem to be the same kind of differences. So it looks like the premise of the test is wrong? I'm a bit confused by this...

jodosha: @nerdinand Thanks for getting back. Did you made sure that the following lines were ported to RSpec? https://github.com/hanami/view/blob/a04b95260b251a36e19de0ec8a29a9aa7a309852/test/test_helper.rb#L42-L56

They seem to solve this equality problem of the configuration **and** the issue about `LoadPaths#include?` mentioned in the description of this PR.

nerdinand:
@jodosha I'm glad to report that everything is fixed now! Things aren't running on Travis yet, I'll look into that, but locally all the specs pass and my hacks are removed.

nerdinand:
@jodosha Turns out Travis was using an ancient version of slim (`0.6.1`) from the bundle cache to do the build. I changed the Gemfile to require `~> 3.0.0`, now it works!

PR #133, opened 2017-08-30
Introduce Hanami::Layout#local

jodosha:
To explain the need of this enhancement let me go straight with an example:

```ruby
# apps/web/views/application_layout.rb
module Web
  module Views
    class ApplicationLayout
      include Web::Layout

      def render_flash
        return if local(:flash).nil?

        local(:flash).map do |type, message|
          html.div(message, class: "flash-#{type}")
        end.join("\n")
      end
    end
  end
end
```

To be used like this:

```erb
# apps/web/templates/application.html.erb
<%= render_flash %>
```

---

Until now, `#local` was only accessible from the associated template. So we were forced to write the code only in this way:

```erb
# apps/web/templates/application.html.erb
<% unless local(:flash).nil? %>
  <% local(:flash).each do |type, message| %>
    <div class="flash-<%= type %>">
      <%= message %>
    </div>
  <% end %>
<% end %>
```

---

To me it was surprising we added `#local` only to the template, but not to the layout. I think there should be a parity between them.

### DIFF
--- a/lib/hanami/layout.rb
+++ b/lib/hanami/layout.rb
@@ -142,7 +142,7 @@ module Hanami
     # It tries to invoke a method for the view or a local for the given key.
     # If the lookup fails, it returns a null object.
     #
-    # @return [Objeect,Hanami::View::Rendering::NullLocal] the returning value
+    # @return [Object,Hanami::View::Rendering::NullLocal] the returning value
     #
     # @since 1.1.0
     #
--- a/lib/hanami/layout.rb
+++ b/lib/hanami/layout.rb
@@ -139,6 +139,27 @@ module Hanami
       template.render(@scope, &Proc.new{@rendered})
     end
 
+    # It tries to invoke a method for the view or a local for the given key.
+    # If the lookup fails, it returns a null object.
+    #
+    # @return [Objeect,Hanami::View::Rendering::NullLocal] the returning value
+    #
+    # @since 1.1.0
+    #
+    # @example Safe method navigation
+    #   class ApplicationLayout
+    #     include Hanami::Layout
+    #
+    #     def render_flash
+    #       return if local(:flash).nil?
+    #
+    #       # ...
+    #     end
+    #   end
+    def local(key)
+      @scope.local(key)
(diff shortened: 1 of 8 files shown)


lib/hanami/layout.rb
```diff
@@ -139,6 +139,27 @@ def render
       template.render(@scope, &Proc.new{@rendered})
     end
 
+    # It tries to invoke a method for the view or a local for the given key.
+    # If the lookup fails, it returns a null object.
+    #
+    # @return [Objeect,Hanami::View::Rendering::NullLocal] the returning value
```
AlfonsoUceda: s/Objeect/Object

jodosha:
@AlfonsoUceda Thanks for the quick review. Yes, this should be documented.

glaszig:
the tests for this only work because `local :flash` returns a `Hash` from the fake exposure that responds to `map`. in reality, you get a `Hanami::Action::Flash` which has no hash-like interface.

i'm using this patch for now:

```rb
class Hanami::Action::Flash
  def map &blk
    last_request_flash.map &blk
  end
end
```

mereghost:
@jodosha @AlfonsoUceda Based on what @glaszig said, wouldn't making `Hanami::Action::Flash` quack as a `Hash` the next step?

jodosha:
@mereghost 
> wouldn't making `Hanami::Action::Flash` quack as a `Hash` the next step?

Nope. I have strong opinions against it.
https://tenderlovemaking.com/2014/06/02/yagni-methods-are-killing-me.html

---

What we can do is to implement only `Flash#map` and `#each` in `hanami-controller`.

jodosha:
@glaszig Thanks for reporting this. It was a naive testing approach from my side. The point of this PR is to let layouts to access locals as objects in general. But yeah, `Flash` needs `#map` and `#each`.

jodosha:
@glaszig We added `Flash#each` and `#map`: https://github.com/hanami/controller/pull/251

They will be available in Hanami v1.2.0, released next week.

Issue #139, opened 2017-11-09
About partial template and local overwrite


cveneziani:
I've encountered two behaviours that I didn't expect regarding locals and partial templates.

## Exposure overwrite

When rendering a partial template with a local, it overwrites my exposure that has the same name as the partial local.

Here is an example:

```ruby
# apps/web/controllers/home/show.rb
module Web::Controllers::Home
  class Show
    include Web::Action

    expose :name

    def call(params)
      @name = 'Hanami'
    end
  end
end
```

```erb
<!-- apps/web/templates/home/show.html.erb -->
<h1>Home page</h1>

<%= name %> <!-- Hanami -->
<%= render partial: 'some_partial', locals: { name: 'Partial' } %>
<%= name %> <!-- Partial -->
```

Then it means that I can't rely anymore on the exposed `name`. 

## Partial local available in calling template

```ruby
# apps/web/controllers/home/show.rb
module Web::Controllers::Home
  class Show
    include Web::Action

    # nothing exposed

    def call(params)
    end
  end
end
```

```erb
<!-- apps/web/templates/home/show.html.erb -->
<h1>Home page</h1>

<%= render partial: 'some_partial', locals: { name: 'Partial' } %>
<%= name %> <!-- Partial -->
```

I didn't expect to have access from my template to a local defined for a specific partial template.

Are these two use cases intended to work this way?

AlfonsoUceda:
@cveneziani thanks for spotting this problem, I'll take care of this.

jodosha:
@cveneziani Thanks for the report.

---

The first case ("Exposure overwrite") works as intended: if you specify an inline override it takes the precedence over the broader scope value. Conceptually it's like:

```ruby
foo = "hello"

def call
  foo = "world"
  puts foo
end

call
# => "world"
```

---

For the second case ("Partial local available in calling template") I think that is a problem that should be fixed.

cveneziani:
I agree that in the context of the partial the inline override takes precedence over the broader scope value. My problem is later on, my initial template gets the overwritten value.

If I take back your example, I expect to have this behavior:

```ruby
foo = "hello"

def call
  foo = "world"
  puts foo
end

call
# => "world"
foo 
# => "hello"
```

But in fact, I get that: 

```ruby
foo = "hello"

def call
  foo = "world"
  puts foo
end

call
# => "world"
foo 
# => "world"
```

Is that the intended behavior?

jodosha:
> Is that the intended behavior?

[...]

PR #140, opened 2017-11-13
Fix overwrite local in partials

AlfonsoUceda:
@cveneziani can you check that #139 is fixed with this PR? thanks!


AlfonsoUceda:
thanks @cveneziani very weird because it was tested here: https://github.com/hanami/view/pull/140/files#diff-e788605daa4d58de3e86e5963ca19c25R18

I'll take a look anyway :)

cveneziani:
Not sure this test takes into account when the view presents the exposure :)

AlfonsoUceda:
nop, I'll try to reproduce it from hanami, because I'd need to create a test inside hanami.

AlfonsoUceda:
@cveneziani I tried with the new fix and it worked, can you try it again please? Thanks!

@jodosha can you take a look at this PR?

cveneziani:
It's working **but** it breaks something else. This time, it might be due to a misuse of the view on my side.

**Here is the new use case:** sharing view code and using it directly within a partial template

Source: http://hanamirb.org/guides/1.1/views/share-code/#prepare

## The shared part

```ruby
# apps/web/application.rb
controller.prepare do
  include Web::SetCurrentUser
end

view.prepare do
  include Hanami::Helpers
  include Web::Assets::Helpers
  
  include Web::Views::PresentCurrentUser
end
```

```ruby
# apps/web/controllers/set_current_user.rb
module Web
  module Users
    module SetCurrentUser
      def self.included(action)
        action.class_eval do
          expose :current_user
        end
      end

      def current_user
        @current_user ||= UserRepository.new.find(....)
      end
    end
  end
end
```

```ruby
# apps/web/views/present_current_user.rb
module Web::Views
  module PresentCurrentUser
    def current_user
      return unless locals[:current_user]
      UserPresenter.new(locals[:current_user])
    end
  end
end
```

```ruby
# apps/web/presenters/user_presenter.rb
class UserPresenter
  include Hanami::Presenter

  def avatar_url
    # Computing some url
    return 'http://hanamirb.org/images/logo.png'
  end
end
```

## Using the shared parts

```ruby
# apps/web/controllers/home/show.rb
module Web::Controllers::Home
  class Show
    include Web::Action
    # SetCurrentUser already included by default

    # nothing exposed

    def call(params)
    end
  end
end
```

```erb
<!-- apps/web/templates/home/show.html.erb -->
<h1>Home page</h1>

<img src="<%= current_user.avatar_url %>"><!-- nice image of http://hanamirb.org/images/logo.png -->
<%= render partial: 'some_partial' %>
```

```erb
<!-- apps/web/templates/home/_some_partial.html.erb -->
<img src="<%= current_user.avatar_url %>"><!-- Crashes -->
```

Previously this code used to work. The partial had access to the "global" presented `current_user`.

With the new changes, now I get the following error: `NoMethodError: undefined method avatar_url for #<User:0x...>`. The partial has access to the global `current_user` but not to the presented current user of course.

In order to fix that, I just have to `render partial: "some_partial", locals: { current_user: current_user }`.

Then I'm wondering. 
Is this the way we should do: passing the presented current user again to the partial template? 
Or since the partial template already has access to the global current user, should it has access to the presented current user instead?

AlfonsoUceda:
@cveneziani thanks for the detailed answer, let me fix it ;)

AlfonsoUceda:
@cveneziani I think I understand your problem and it seems it is an error in your code, can you try remove `return unless locals[:current_user]` from `PresentCurrentUser`? The problem is because when you are inside the partial, it applies another scope and it doesn't have `locals[:current_user]` so it crashes because `locals` is evaluated in that scope.

cc @jodosha

cveneziani:
I've tried to remove `return unless locals[:current_user]` from `PresentCurrentUser` and I have the same result: `NoMethodError: undefined method avatar_url for #<User:0x...>`

From what I see, the partial has access to `current_user` but it's the one provided by the action, not the one exposed by the view.

My example was way too complicated, sorry for that.

If I take back the first example, here is the use case:

```ruby
# apps/web/controllers/home/show.rb
module Web::Controllers::Home
  class Show
    include Web::Action

    expose :name

    def call(params)
      @name = 'Hanami'
    end
  end
end
```

```ruby
# apps/web/views/home/show.rb
module Web::Views::Home
  class Show
    include Web::View

    def name
      "Presented #{locals[:name]}"
    end
  end
end
```

```erb
<!-- apps/web/templates/home/show.html.erb -->
<h1>Home page</h1>

<%= name %> <!-- Presented Hanami -->
<%= render partial: 'some_partial' %> <!-- Hanami -->
<%= name %> <!-- Presented Hanami -->
```

This time, the partial doesn't overwrite `name`. 

But then, should the partial have access to the exposed value (from the action) or to the presented value (from the view)?

AlfonsoUceda:
@cveneziani don't worry hehe, but yes, the partial should have the same access like the template. I'll look into it.

cveneziani:
> the partial should have the same access like the template

If and only if the partial doesn't overwrite the presented value, right?

Without overwrite:

```erb
<!-- apps/web/templates/home/show.html.erb -->
<h1>Home page</h1>

<%= name %> <!-- Presented Hanami -->
<%= render partial: 'some_partial' %> <!-- Presented Hanami -->
<%= name %> <!-- Presented Hanami -->
```

With overwrite:

```erb
<!-- apps/web/templates/home/show.html.erb -->
<h1>Home page</h1>

<%= name %> <!-- Presented Hanami -->
<%= render partial: 'some_partial', locals: { name: 'Partial' } %> <!--  Partial -->
<%= name %> <!-- Presented Hanami -->
```

Thanks again for your time.

AlfonsoUceda:
@cveneziani exactly ;)

AlfonsoUceda:
@cveneziani I think I've fixed the problem but I need to clean the code ;)

cveneziani:
That's great news. Thanks :)

AlfonsoUceda:
@cveneziani this is done, it was tricky because I needed to refactor :) can you check again?

AlfonsoUceda:
I tested this branch with hanami: https://travis-ci.org/hanami/hanami/builds/307111073

AlfonsoUceda:
@hanami/core this PR is ready for review.

jodosha:
@cveneziani Can you please try this PR with your project and tell us if it works for you? Thanks in advance! 😄

AlfonsoUceda:
@jodosha thanks for the refactoring ;(

cveneziani:
Ok so I've checked and I confirm, it's working well with "regular" templates but I have an unexpected behavior with partials rendered from the **layout template**.

Here is what I have:

```ruby
# apps/web/controllers/home/show.rb
module Web::Controllers::Home
  class Show
    include Web::Action

    expose :name

    def call(params)
      @name = 'Hanami'
    end
  end
end
```

```ruby
# apps/web/views/home/show.rb
module Web::Views::Home
  class Show
    include Web::View

    def name
      "Presented #{locals[:name]}"
    end
  end
end
```

```erb
<!-- apps/web/templates/layouts/application.html.erb -->
<!DOCTYPE html>
<html>
  <head>
    <title>Hanami app</title>
  </head>
  <body>
    <%= name %> <!-- Presented Hanami -->
    <%= render partial: 'some_partial' %> <!-- Hanami -->
    <%= yield %>
  </body>
</html>
```

```erb
<!-- apps/web/templates/home/show.html.erb -->
<h1>Home page</h1>

<%= name %> <!-- Presented Hanami -->
<%= render partial: 'some_partial' %> <!-- Presented Hanami -->
<%= name %> <!-- Presented Hanami -->
```

The partial rendered from the layout used to have access to the presented `name` (eg. 'Presented Hanami'). It's not the case anymore.

AlfonsoUceda:
@cveneziani 😅 thanks, I think how can I fix it ;)

AlfonsoUceda:
@cveneziani I've reproduced that error and fix it ;) sorry for the noise but thank you very much.

lib/hanami/view/rendering/layout_scope.rb
```diff
@@ -230,7 +230,9 @@ def method_missing(m, *args, &blk)
           # that we want to be frozen in the future
           #
           # See https://github.com/hanami/view/issues/130#issuecomment-319326236
-          if @scope.respond_to?(m, true)
+          if @scope && @scope.respond_to?(m, true) && @scope.locals.has_key?(m) && layout.respond_to?(m)
```
jodosha: Is there any reason why we don't use `layout.respond_to?(m, true)`?

AlfonsoUceda: there isn't, I just copy&paste from this line https://github.com/hanami/view/pull/140/files/bd867cee6c67fa9199390ee5c613071dfb321202..79d76e8af4d722a27bf2c938973134f75233f0a5#diff-5d90b9ac33da24fadae41a69b3800763R237

jodosha: Please verify that private methods work as well. Thanks 😄

cveneziani:
I tested the last commits and I have two issues:

- Rendering layout partial from another layout partial raises an error
- Exposures not presented are not available anymore from a layout partial

Here is what I have:

```ruby
# apps/web/controllers/home/show.rb
module Web::Controllers::Home
  class Show
    include Web::Action

    expose :name

    def call(params)
      @name = 'Hanami'
    end
  end
end
```

```ruby
# apps/web/views/home/show.rb
module Web::Views::Home
  class Show
    include Web::View

    # not presenting name
  end
end
```

```erb
<!-- apps/web/templates/layouts/application.html.erb -->
<!DOCTYPE html>
<html>
  <head>
    <title>Hanami app</title>
  </head>
  <body>
    <%= name %> <!-- Hanami -->
    <%= render partial: 'layouts/some_partial' %> <!-- NameError: undefined local variable or method `name' for #<Hanami::View::Rendering::Subscope:0x00007efbfa5c9a20> -->
    <%= render partial: 'layouts/nesting_partial' %>
    <%= yield %>
  </body>
</html>
```

```erb
<!-- apps/web/templates/layouts/_nesting_partial.html.erb -->
<p>Rendering another partial</p>
<%= render partial: 'layouts/some_partial' %> <!-- NoMethodError: undefined method `to_sym' for nil:NilClass --- view/lib/hanami/view/configuration.rb:445:in `find_partial' -->
```

```erb
<!-- apps/web/templates/home/show.html.erb -->
<h1>Home page</h1>

<%= name %> <!-- Hanami -->
<%= render partial: 'some_partial' %> <!-- Hanami -->
<%= name %> <!-- Hanami -->
```

If I present the `name` again, the layout partial works fine.

AlfonsoUceda:
@cveneziani thanks so much, I'll give it a try later.

cveneziani:
Not sure it's done yet but just in case, with the new commit I still have issues.

AlfonsoUceda:
@cveneziani I know, I made another commit because I needed to fix formats for partials from another folder.

Recap, you pointed out two problems with layouts:
> Exposures not presented are not available anymore from a layout partial

This should be fixed.

> Rendering layout partial from another layout partial raises an error

This should be fixed, in your example, you use the full relative path, please continue using it, I don't know if we will be able to use relative paths to the folder, I think we can't do it.

@jodosha can you take a look?

cveneziani:
I'm sorry for this long delay, got a lot of work :)

So I've tested and everything seems to work really well.
Thanks again for all you patience.

AlfonsoUceda:
@cveneziani thank you very much again :)

PR #145, opened 2018-03-23
Avoid Tilt warning about :disable_escape option

nerdinand:
For HAML templates, this warning is shown for for each loaded template, which is noisy and annoying:
`Haml::TempleEngine: Option :disable_escape is invalid`

Another solution would be to only pass this option if we're dealing with a Slim template...

### DIFF
--- a/lib/hanami/view/template.rb
+++ b/lib/hanami/view/template.rb
@@ -7,11 +7,9 @@ module Hanami
     # @since 0.1.0
     class Template
       def initialize(template, encoding = Encoding::UTF_8)
-        default_options = {default_encoding: encoding}
+        options = {default_encoding: encoding}
 
-        options = default_options
-
-        if Tilt[template] != Tilt::HamlTemplate
+        if File.extname(template) == '.slim'
           # NOTE disable_escape: true is for Slim compatibility
           # See https://github.com/hanami/assets/issues/36
           options.merge!(disable_escape: true)
--- a/lib/hanami/view/template.rb
+++ b/lib/hanami/view/template.rb
@@ -7,9 +7,17 @@ module Hanami
     # @since 0.1.0
     class Template
       def initialize(template, encoding = Encoding::UTF_8)
-        # NOTE disable_escape: true is for Slim compatibility
-        # See https://github.com/hanami/assets/issues/36
-        @_template = Tilt.new(template, nil, default_encoding: encoding, disable_escape: true)
+        default_options = {default_encoding: encoding}
+
+        options = default_options
+
+        if Tilt[template] != Tilt::HamlTemplate
+          # NOTE disable_escape: true is for Slim compatibility
+          # See https://github.com/hanami/assets/issues/36
+          options.merge!(disable_escape: true)
+        end
+
+        @_template = Tilt.new(template, nil, options)
       end
 
       # Returns the format that the template handles.
--- a/spec/support/fixtures/templates/hello_world.html.slim
+++ b/spec/support/fixtures/templates/hello_world.html.slim
@@ -1,0 +1,1 @@
+h1 Hello, World!
--- a/spec/unit/hanami/view/template_spec.rb
+++ b/spec/unit/hanami/view/template_spec.rb
@@ -22,7 +22,7 @@ RSpec.describe Hanami::View::Template do
       end
 
       it 'sets :disable_escape to true with non-haml templates' do
-        template = Hanami::View::Template.new("#{TEMPLATE_ROOT_DIRECTORY}/hello_world.html.erb",)
+        template = Hanami::View::Template.new("#{TEMPLATE_ROOT_DIRECTORY}/hello_world.html.slim")
         expect(template.instance_variable_get(:@_template).__send__(:options)[:disable_escape]).to eq(true)
       end
     end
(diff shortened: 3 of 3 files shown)


davydovanton:
and maybe some specs?

lib/hanami/view/template.rb
```diff
@@ -7,9 +7,17 @@ module View
     # @since 0.1.0
     class Template
       def initialize(template, encoding = Encoding::UTF_8)
-        # NOTE disable_escape: true is for Slim compatibility
-        # See https://github.com/hanami/assets/issues/36
-        @_template = Tilt.new(template, nil, default_encoding: encoding, disable_escape: true)
+        default_options = {default_encoding: encoding}
+
+        options = default_options
```
jodosha: @nerdinand Why is this needed?

lib/hanami/view/template.rb
```diff
@@ -7,9 +7,17 @@ module View
...
+
+        if Tilt[template] != Tilt::HamlTemplate
```
jodosha: This is conceptually wrong: why all the other templates need that `disable_escape` option if it's only for Slim?
Shouldn't it be `if Tilt[template] == Slim::Template`?

jodosha: This will trigger Tilt autoloading of the `HamlTemplate` class, which requires `haml` gem. For applications that don't bundle that gem that will cause a crash.

**With this branch checked out, from a separated directory try to do this:**

```shell
$ pwd
/Users/luca/Desktop/test

$ cat Gemfile
source "https://rubygems.org"

gem "hanami-view", path: "/Users/luca/Code/hanami/view"
```

Then start an IRB session:

```ruby
irb(main):001:0> require "hanami/view"
=> true
irb(main):002:0> Hanami::View::Template.new("index.html.erb")
Traceback (most recent call last):
       16: from /Users/luca/.rubies/ruby-2.5.0/lib/ruby/site_ruby/2.5.0/bundler/vendor/thor/lib/thor/base.rb:466:in `start'
       15: from /Users/luca/.rubies/ruby-2.5.0/lib/ruby/site_ruby/2.5.0/bundler/cli.rb:27:in `dispatch'
       14: from /Users/luca/.rubies/ruby-2.5.0/lib/ruby/site_ruby/2.5.0/bundler/vendor/thor/lib/thor.rb:387:in `dispatch'
       13: from /Users/luca/.rubies/ruby-2.5.0/lib/ruby/site_ruby/2.5.0/bundler/vendor/thor/lib/thor/invocation.rb:126:in `invoke_command'
       12: from /Users/luca/.rubies/ruby-2.5.0/lib/ruby/site_ruby/2.5.0/bundler/vendor/thor/lib/thor/command.rb:27:in `run'
       11: from /Users/luca/.rubies/ruby-2.5.0/lib/ruby/site_ruby/2.5.0/bundler/cli.rb:424:in `exec'
       10: from /Users/luca/.rubies/ruby-2.5.0/lib/ruby/site_ruby/2.5.0/bundler/cli/exec.rb:28:in `run'
        9: from /Users/luca/.rubies/ruby-2.5.0/lib/ruby/site_ruby/2.5.0/bundler/cli/exec.rb:75:in `kernel_load'
        8: from /Users/luca/.rubies/ruby-2.5.0/lib/ruby/site_ruby/2.5.0/bundler/cli/exec.rb:75:in `load'
        7: from /Users/luca/.rubies/ruby-2.5.0/bin/irb:11:in `<top (required)>'
        6: from (irb):2
        5: from (irb):2:in `new'
        4: from /Users/luca/Code/hanami/view/lib/hanami/view/template.rb:14:in `initialize'
        3: from /Users/luca/Code/hanami/view/lib/hanami/view/template.rb:14:in `require'
        2: from /Users/luca/.gem/ruby/2.5.0/gems/tilt-2.0.8/lib/tilt/haml.rb:2:in `<top (required)>'
        1: from /Users/luca/.gem/ruby/2.5.0/gems/tilt-2.0.8/lib/tilt/haml.rb:2:in `require'
LoadError (cannot load such file -- haml)
```

nerdinand: Right, that's another possibility. I just wanted to keep it as close to what we had before as possible...

nerdinand: Right. Won't we have the same problem with `if Tilt[template] == Slim::Template`? Is slim different?

lib/hanami/view/template.rb
nerdinand: I just wanted to make the distinction clearer that the `default_encoding` option is always being passed.

lib/hanami/view/template.rb
jodosha: Yes we would have the same problem, but this is a different perspective from previous comment. The first one is only about boolean logic, this one is a blocker because it would crash apps.

So `if Tilt[template] == Slim::Template` is **NOT** a viable solution.

nerdinand: @jodosha what do you suggest then? Should we do something like `if File.basename(template).include? '.slim'`? Looking at the slim code I see nothing more elegant than that...

nerdinand: @jodosha that would only look at the last extension, I guess that's fine in this case...

nerdinand:
@jodosha is it ok like this?

PR #146, opened 2018-04-09
Ensure to safely use partial locals that can't be duplicated

### LINKED ISSUE #139: About partial template and local overwrite

jodosha:
With #140 we introduced deep duplication of locals to fix #139.

But objects that can't be duplicated (e.g. `ROM::Relation::Composite`, aka `ROM::Repository::RelationProxy`), will raise a `NoMethodError`.

Here's a typical scenario:

```ruby
class ContributorRepository < Hanami::Repository
  def listing
    contributors.order { commits_count.desc }
  end
end
```

```ruby
module Web::Controllers::Home
  class Index
    include Web::Action
    expose :contributors

    def call(*)
      @contributors = ContributorRepository.new.listing
    end
  end
end
```

That `contributors` exposure can't be duped, and it raises an error.

### DIFF
--- a/lib/hanami/view/rendering/layout_scope.rb
+++ b/lib/hanami/view/rendering/layout_scope.rb
@@ -126,7 +126,7 @@ module Hanami
         #
         # @since 0.1.0
         def locals
-          Utils::Hash.deep_dup(@locals || @scope.locals)
+          (@locals || @scope.locals).dup
         end
 
         # It tries to invoke a method for the view or a local for the given key.
--- a/lib/hanami/view/rendering/options.rb
+++ b/lib/hanami/view/rendering/options.rb
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require "hanami/utils/hash"
-
 module Hanami
   module View
     module Rendering
@@ -13,7 +11,7 @@ module Hanami
         # @since 1.1.1
         # @api private
         def self.build(options, locals, format)
-          Utils::Hash.deep_dup(options).tap do |opts|
+          options.dup.tap do |opts|
             opts[:format] = format
             opts[:locals] = locals
             opts[:locals].merge!(options.fetch(:locals) { ::Hash.new }).merge!(format: format)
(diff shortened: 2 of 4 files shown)


jodosha:
@cveneziani Hi, would you mind to check if this patch doesn't break your app? TYVM.

AlfonsoUceda:
@jodosha good catch 👍 have you checked this branch in a hanami branch? just to check the CI.

jodosha:
@AlfonsoUceda I personally ran into this problem, while updating `jodosha/hall_of_fame` to the the latest rc2.

With this branch, the app doesn't crash anymore.

jodosha:
I have to merge this without @cveneziani confirmation, but only with mine.
This PR is blocking my work of 1.2.0 release 😢

cveneziani:
Hello, sorry for the late reply. I've tested all the use cases that I could remember and it's working fine.

jodosha:
@cveneziani No need to be sorry at all. It's me that this time was in a hurry. I'm actually sorry.
Glad that it worked. 😄

cveneziani:
Hehe no worries then :)

Issue #151, opened 2018-08-22
raw view helper does not work


papile:
In the docs there is reference to the view helper `#raw` to prevent HTML escaping on contents in the view. Using this in a view results in it being delegated to `method_missing` which only takes one parameter, the method in views/render.rb. This results in an argument error 2 for 1. Upon looking at the specs I see `_raw` which when used in the view works as expected.

I am not sure if it is that the docs are wrong, or if `raw` should be picked up in the DSL and mapped to `_raw`

Documentation that I am referring to: http://hanamirb.org/guides/1.2/views/mime-types/#view-for-specific-format [](url) I am using Hanami 1.2, Ruby 2.5.1

papile:
`module Api::Views::Items
  class Show
    include Api::View

    def render
      #broken "wrong number of arguments (given 2, expected 1)"
      raw 'hello'
      #works
      _raw 'hello'
    end
  end
end`

I have updated the docs and can submit a PR if it is supposed to be `_raw` I just am new to this project and didn't know the intention.

jodosha:
@papile Are you using `hanami-view` in isolation? If so, please note that `#raw` it isn't defined in this gem, but `#_raw` is.

---

`#raw` is a method from `hanami-helpers`, so it's available in the view context only if you use `hanami`.

Please double check that `apps/api/application.rb` has the following lines:

```ruby
view.prepare do
  include Hanami::Helpers
end
```

If I comment it, then I'm able to reproduce the problem.

papile:
Yes I was using it in Hanami. My `application.rb` has the include helpers block. I am not sure what the difference is between my app and the simple bookshelf one. So I suppose my view did not have the helpers included for some reason despite it having included `Api::Views` and having the helpers included in application.

One thing I've noticed with Hanami is with the large amount of metaprogramming, strange things happen sometimes depending on naming, sequence and context.  Also if you get something slightly wrong, you get errors and behaviors that are hard to debug with blocks being instance/class eval'ed in dynamically generated classes. So it working in `Bookshelf` indicates there is something happening in my app specifically but as a bug reporter I have to now find out what it is :)

jodosha:
@papile You can also do `Api::Views::Items::Show.included_modules` to check if `Hanami::Helpers` is included.

PR #155, opened 2018-09-20
Ensure to raise NoMethodError when a missing error with arity > 0 is used

jodosha:
While trying to reproduce https://github.com/hanami/view/issues/151 I discovered a problem.

If we reference a missing method with arity > 0, then `Hanami::View::Rendering#method_missing` crashes because the signature is: `method_missing(m)`. To override this Ruby method, we should accept any argument.

It ended up to raise an error like this:

```ruby
ArgumentError at 
wrong number of arguments (given 2, expected 1)
```

When we use an unknown method it should raise `NoMethodError` instead.
This PR addresses the problem.


AlfonsoUceda:
LGTM @jodosha 👍

Issue #153, opened 2018-08-28
Undefined method `locals'


steakknife:
Maybe I'm holding it wrong, but it looks like an API changed or something didn't load properly. Suggestions or is this a bug?

### MCE

```
gem install hanami -v 1.2.0
gem install haml -v 5.0.4
hanami new Bookstore --database=postgres --test=rspec --template=haml
HANAMI_ENV=test bundle exec hanami db create
rake spec
```

### Output

```
Failures:

  1) Web::Views::ApplicationLayout contains application name
     Failure/Error: let(:rendered) { layout.render }

     NoMethodError:
       undefined method `locals' for #<Hanami::View::Template:0x00007fd3ce6adfc0>
     # ~/.gem/ruby/2.5.1/gems/hanami-view-1.2.0/lib/hanami/view/rendering/layout_scope.rb:233:in `method_missing'
     # ~/.gem/ruby/2.5.1/gems/tilt-2.0.8/lib/tilt/haml.rb:23:in `evaluate'
     # ~/.gem/ruby/2.5.1/gems/tilt-2.0.8/lib/tilt/template.rb:109:in `render'
     # ~/.gem/ruby/2.5.1/gems/hanami-view-1.2.0/lib/hanami/view/template.rb:41:in `render'
     # ~/.gem/ruby/2.5.1/gems/hanami-view-1.2.0/lib/hanami/layout.rb:139:in `render'
     # ./spec/web/views/application_layout_spec.rb:5:in `block (2 levels) in <top (required)>'
     # ./spec/web/views/application_layout_spec.rb:9:in `block (2 levels) in <top (required)>'

Finished in 0.00574 seconds (files took 2.33 seconds to load)
1 example, 1 failure
```

### Versions

- macOS 10.13.6 (17G65)
- ruby 2.5.1
- gem 2.7.7

#### Gem Versions

haml (5.0.4)
haml-lint (0.999.999)
haml-rails (1.0.0)
haml_lint (0.28.0)
hanami (1.2.0)
hanami-assets (1.2.0)
hanami-cli (0.2.0)
hanami-controller (1.2.0)
hanami-helpers (1.2.2)
hanami-mailer (1.2.0)
hanami-model (1.2.0)
hanami-router (1.2.0)
hanami-utils (1.2.0)
hanami-validations (1.2.2)
hanami-view (1.2.0)
hanami-webconsole (0.1.0)
[...]

PR #156, opened 2018-10-03
Allow layout to be rendered directly for testing purposes

jodosha:
Until Hanami v1.2.0 we generate a wrong spec for layout unit tests.

```ruby
# spec/web/views/application_layout_spec.rb
require "spec_helper"

RSpec.describe Web::Views::ApplicationLayout, type: :view do
  let(:layout)   { Web::Views::ApplicationLayout.new(template, {}) }
  let(:rendered) { layout.render }
  let(:template) { Hanami::View::Template.new('apps/web/templates/application.html.haml') }

  it 'contains application name' do
    expect(rendered).to include('Web')
  end
end
```

If we look at the signature of `Hanami::Layout#initialize` we'll see it accept two arguments: `scope` and `rendered`.

The first must be a `Hanami::View::Rendering::Scope` instance. It's an internal object composed by `hanami-view`, that wraps the locals and the **format** to be rendered (e.g. `:html`).

The second argument is a string, and it's the result of the view rendering. E.g. a template is made of:

```erb
<p>Hello <%= name %>!</p>
```

Then the `rendered` value will be `<p>Hello Luca!</p>`.

---

Getting back to our generated code above, it **doesn't respect the signature** of `Hanami::Layout#initialize`. With ERB templates it **accidentally worked**, but for HAML, it didn't (see https://github.com/hanami/view/issues/153).

---

The purpose of this PR is to make `Hanami::Layout` ready to be unit tested properly with:

```ruby
rendered = ApplicationLayout.new({ format: :html }, "contents").render
```

On a subsequent PR, I'll fix the code generator for `hanami`.

---

Fixes https://github.com/hanami/view/issues/153

### DIFF
--- a/lib/hanami/layout.rb
+++ b/lib/hanami/layout.rb
@@ -2,6 +2,7 @@ require 'hanami/utils/class_attribute'
 require 'hanami/view/rendering/layout_registry'
 require 'hanami/view/rendering/layout_scope'
 require 'hanami/view/rendering/null_layout'
+require 'hanami/view/rendering/null_view'
 
 module Hanami
   # Layout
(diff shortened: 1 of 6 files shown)


AlfonsoUceda:
after talking some details with Luca, LGTM 👍

PR #163, opened 2019-09-25
Add support for passing blocks to LayoutScope#render (fixes #141)

### LINKED ISSUE #141: Partial templates that can yield
Hi,

I've had this requirement from time to time, I don't know if there is a specific name for it. I have a big chunk of HTML that I use on different pages, but I want to be able to customise something inside it. So I'm going to use small examples that aren't worth the trouble, but in practice it can be large templates, and the nesting makes sense. I'm using Slim:

```ruby
div.container
  = render partial: './my_partial' do
    | This is the content for my page #1
```

And the partial goes like this:
```ruby
nav
  a.nav-item
  a.nav-item
div.decoration
  = yield
```

This is currently not working.

I hacked into the code, changed two lines and now it works, cf 7ad6120ba9b26f50d6d36dc5ebff990e9e794700 . But I have no idea why it was not possible in the first place, so maybe it's not supposed to work this way. And I also don't know how to write a spec for that use case, so I'd like further input before doing any kind of pull request.

unleashy:
#141 got closed, but I really wanted this so I went ahead and implemented it. I'm afraid the specs aren't very up to par (there's only one and for some reason I feel like this is gonna be buggy) because I honestly couldn't understand how exactly they worked (like, why is the integration folder disabled?) so I would honestly like some input on this. Thanks!


unleashy:
Thank you :heart_decoration: [“Componentised” partials](https://evilmartians.com/chronicles/evil-front-part-1) are exactly the use case for this—I used them in Rails and it was great!

> Can you please explain why do you think the spec is buggy? We can try to reproduce real usage that you have in mind for your app, so a more realistic spec is better for the purpose.

I'm just afraid that it's not gonna test everything (since it's just one spec), but looking back I think it's fine. There's not much to go wrong here (hopefully).

> Integration specs aren't disabled, how do you run specs? The expected way to run the test suite is to use `script/ci` from this repo.

Ahh, I was using `bundle exec rake` :stuck_out_tongue_closed_eyes: I wasn’t aware of `script/ci`, sorry.

jodosha:
@unleashy Thank you! I'm gonna add more integration tests, inspired by @evilmartians articles.

PR #166, opened 2019-12-30
Hanami::View::Rendering::TemplateName: Fix frozen string literal error

MITSUBOSHI:
## Problems
It's impossible to run `hanami server` when using ruby 2.7.0 (the reasons are below)

### Causes
- https://github.com/hanami/view/blob/9af8499ae1e7ce81c7eb85167222d9b3bfff2ec7/lib/hanami/view/rendering/template_name.rb#L45
  - It's using bang method `gsub!` here. It works as long as `@name` as String is not frozen.
- https://github.com/hanami/view/blob/9af8499ae1e7ce81c7eb85167222d9b3bfff2ec7/lib/hanami/view/dsl.rb#L208
  - It passes `Module#name` method as arugument. And it will be assigned to `@name` in `Rendering::TemplateName#initialize`.
- The behavioir `Module#name` has been changed since Ruby 2.7.0. *[ref](https://rubyreferences.github.io/rubychanges/2.7.html#core-methods-returning-frozen-strings)
  - the previous behavior: `(Foo = Module.new).name.frozen? #=> false`
  - the current behavior: `(Foo = Module.new).name.frozen? #=> true`

## Execution Environment
- ruby 2.7.0
- hanami 1.3.3
- hanami-view 1.3.2

## Steps to reproduce the problem
1. Use ruby `2.7.0` 
2. `hanami new bookshelf` & `bundle install`
3. `bundle exec hanami server` -> then, the issued problem occurs
 
## others
<details>
<summary>error log</summary>

```log
Boot Error
Something went wrong while loading /Users/mitsubosh/Practice/043_hanami/bookshelf/config.ru

FrozenError: can't modify frozen String: "Web::Views::ApplicationLayout"
/Users/mitsubosh/Practice/043_hanami/bookshelf/vendor/bundle/ruby/2.7.0/gems/hanami-view-1.3.2/lib/hanami/view/rendering/template_name.rb:45:in `gsub!'
/Users/mitsubosh/Practice/043_hanami/bookshelf/vendor/bundle/ruby/2.7.0/gems/hanami-view-1.3.2/lib/hanami/view/rendering/template_name.rb:45:in `replace!'
/Users/mitsubosh/Practice/043_hanami/bookshelf/vendor/bundle/ruby/2.7.0/gems/hanami-view-1.3.2/lib/hanami/view/rendering/template_name.rb:30:in `block in compile!'
[...]

PR #167, opened 2019-12-30
fix travis

MITSUBOSHI:


MITSUBOSHI:
@jodosha 
It seems that Travis CI in this repo doesn't work correctly.
It fails because of using the latest version `bundler` with older ruby and this issue #168.

So, I fixed. Could you please review this PR?
(I would like you to review another PR https://github.com/hanami/view/pull/166, also :) )

PR #172, opened 2020-03-18
Add Hanami application integration

timriley:
This is a work-in-progress.


timriley:
App integration is now provided by `lib/extensions/view.rb` in hanami.

PR #173, opened 2020-05-12
Clean Hanami application integration

timriley:
This PR implements the cleanest possible Hanami application integration for Hanami views.

Our first pass at having hanami views integrate with the application involved us having to ask users to inherit from Hanami::View in a slightly strange way:

```ruby
module Main
  class View < Hanami::View[:main] # <- this bit
  end
end
```

This worked well enough, but having to specify you want a view for the :main slice when you're already inside the Main slice's own namespace felt a little redundant.

Further, the extra functionality for these "application views" was inserted by the hanami gem itself, by monkey-patching the Hanami::View class, which didn't feel clean either.

This PR addresses both these issues, using the new Hanami application features available in https://github.com/hanami/hanami/pull/1052.

You can now define views with an Hanami app like this:

```ruby
module Main
  class View < Hanami::View # <- Look, no more boilerplate!
  end
end
```

So when you inherit directly `Hanami::View` now, we check to see:

- If an Hanami application exists
- And if the subclass (`Main::View` in this example) belongs to the application and/or one of its slices, then it dynamically includes an `ApplicationView` module build specifically for that application or slice. This module configures the view for the application.

To make this work cleanly, we move the standard view behaviour into a `StandaloneView` module, which means `ApplicationView` has the opportunity to introduce new/specialised behaviour, because it will be included earlier in the ancestors chain, so it can use `super` in the standard way and also not have to worry about having to clobber methods directly defined on `Hanami::View`.

The ancestors for an integrated view will now look like this, more or less:

`[Main::View, #<ApplicationView>, StandaloneView, ...`

This strategy has proven to work very nicely, and it's one I'll follow for the integration of Hanami actions into the full application (which requires more new behaviour to be added compared to what we do with views).


timriley:
Ready for review @jodosha! Ignore the red build flags, those seem to be for services that the previous hanami-view was configured to use, not the code we've imported from dry-view.

I'm also flagging @solnic and @flash-gordon here because I'd love to hear if you have any thoughts about the way I've structured the code here for having `Hanami::View` behave slightly differently depending on whether it is subclasses within an Hanami app (i.e. moving the existing logic to `StandaloneView`, then adding that `.inherited` hook and using the module builder pattern to bring in a specific `ApplicationView`).

.github/workflows/ci.yml
```diff
@@ -29,7 +29,6 @@ jobs:
           - "2.7"
           - "2.6"
           - "2.5"
-          - "2.4"
```
solnic, line 32: @timriley are we officially dropping support for 2.4? if yes, then it shouldn't be part of this PR

timriley: @solnic I had to do this because hanami-utils unstable branch would no longer install on Ruby 2.4, so that particular build would fail.

@jodosha — what's your perspective on how to sort this out? And what's our Ruby version support for these unstable/eventually-2.0 branches?

timriley: For those following this PR: we've sorted this out. Dropping 2.4 (at least) from unstable branches of gems.

PR #177, opened 2020-07-16
Create ApplicationContext and (initial) context helpers for use within full Hanami apps

timriley:
This PR adds the beginning of Seamless Application Integration™️ to `Hanami::View::Context`.

Like we already do for `Hanami::View` and `Hanami::Action`, now when you inherit from `Hanami::View::Context` from within an Hanami application module, you'll get extra application-specific behaviours provided automatically.

To start with, this is reader methods for the `#request`, `#session`, and `#flash` (which will work with the `request` and `response` objects already provided as part of the `#view_context_options` in `Hanami::Action`'s view rendering integration.

No fancy metaprogramming required for this one, just a standard module with some instance methods.

n.b. this is certainly _not_ the final endpoint of the application context behavior, but it's a helpful starting point and required for application authors to fully access the features provided via the action/view rendering integration.

Along with this, this PR also provides an experimental new `Hanami::View::ContextHelpers::ContentHelpers` module, which can be explicitly included into an application's context class to opt into some extra behavior on the application's view context. I imagine this pattern could be extended to provide an entire "standard library" of opt-in view context behavior. If this proves a useful approach, I'd imagine we could add some class-level niceties to make it easier to opt in without having to provide the fully qualified module names, but in the meantime, this is a simpler way to test the waters.

### DIFF
--- a/lib/hanami/view/application_context.rb
+++ b/lib/hanami/view/application_context.rb
@@ -3,8 +3,8 @@
 module Hanami
   class View
     module ApplicationContext
-      def initialize(**options)
-        @inflector = options.fetch(:inflector) { Hanami.application.inflector }
+      def initialize(inflector: Hanami.application.inflector, **options)
+        @inflector = inflector
         super
       end
 
--- a/lib/hanami/view/application_context.rb
+++ b/lib/hanami/view/application_context.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+require_relative "context"
+
+module Hanami
+  module View
+    class ApplicationContext < Context
+      attr_reader :inflector
+
+      def initialize(**options)
+        @inflector = options.fetch(:inflector) { Hanami.application.inflector }
+        super(**options)
+      end
+
+      def request
+        _options.fetch(:request)
+      end
+
+      def session
+        request.session
+      end
+
+      def flash
+        response.flash
+      end
+
+      private
+
+      # TODO: put `flash` on Request and stop passing response to view context
+      def response
+        _options.fetch(:response)
+      end
+    end
+  end
+end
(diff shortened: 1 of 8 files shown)


lib/hanami/view/context_helpers/content_helpers.rb
```diff
@@ -0,0 +1,26 @@
+module Hanami
```
jodosha, line 1: @timriley This is fine to put as a proof of concept here, but we should consider to use `hanami-helpers` in the future. 🙂

lib/hanami/view/application_context.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+module Hanami
+  class View
+    module ApplicationContext
+      def initialize(**options)
+        @inflector = options.fetch(:inflector) { Hanami.application.inflector }
```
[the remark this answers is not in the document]
timriley: @jodosha You're right, that approach wasn't necessary! I think I did it to be consistent with the params signature of the base `Context#initialize`:

```ruby
def initialize(render_env: nil, **options)
```

But this totally works and creates a more explicit signature:

```ruby
def initialize(inflector: Hanami.application.inflector, **options)
  @inflector = inflector
  super
end
```

I'm pushing that change up now.

BTW, we actually need to pass _all_ kwargs up to the base initializer so it can keep them all in `@_options` and restore them when we prepare the view context for a given render environment:

```
# From View::Context:

      # @api private
      def for_render_env(render_env)
        return self if render_env == _render_env

        self.class.new(**_options.merge(render_env: render_env))
      end
```

This is why we need to call `super` (which ensures the inflector is passed up) instead of `super(**options)` which will not pass the inflector.

lib/hanami/view/context_helpers/content_helpers.rb
timriley: @jodosha haha, I forgot about this package! Yep, I'd be happy to move this work into there when we get to the point of expanding this beyond a single-file experiment.

PR #178, opened 2020-09-26
Provide context object from template to layout

timriley:
This makes it possible for the layout to access any context state changed from within the template, which is necessary to enable `content_for`-style behaviour, or e.g. for context helpers that allow the templates to set a page title, which is then output from the layout.


timriley:
This is already taken care of by changes in the last couple years.

PR #179, opened 2020-09-26
Use effects to access the render env, as well as locals from context

timriley:
In this PR we're achieving two things using dry-effects (the first I did [way back in September 2020](https://timriley.info/writing/2020/10/06/open-source-status-update-september-2020/), and the second I'm adding just now):

## Access to the render env

This removes the cyclic dependency between a render environment and its context, part builder, and scope builder.

It also means we don’t need to create new copies of the context, part builder, and scope for each render environment we create (typically, there are 3 environments for a given rendering: a base environment, then a template- and layout-specific environment).

This is a particularly helpful in the case of the context object, because application authors may want to set state in the context object from within the template rendering, to make it available when the layout renders. This was not possible to do until now because both the template and layout renderings had different context objects.

## Access to locals from context

This is a significant change to the purpose and function of the context object - now you can access the _current_ locals via a `#locals` reader effect.

This should enable us to create more streamlined helper APIs that require implicit access to locals

## TODO

- [x] Add a test that relies upon an identical context object being used in both template and layout renderings
- [x] Remove `Context#_options` and `#with`
- [x] Work out whether we need Part#new anymore (I think no)
- [ ] We've disabled access to the context from within exposure blocks. Decide whether this is OK or needs to be reinstated somehow.
- [ ] Rescue from `Dry::Effects::Errors::MissingStateError` for our `render_env` and raise a more helpful error in its place
- [ ] (Later) Create some hanami-view-specific testing helpers so app authors don't need to include a dry effects handler directly in their tests


lib/hanami/view/context.rb
```diff
@@ -13,67 +14,12 @@ class View
...
-      #   my_context_with_request = my_context.with(request: request)
-      #
-      # @api public
-      def with(**new_options)
-        self.class.new(
-          render_env: _render_env,
-          **_options.merge(new_options)
-        )
+      def locals
+        scope._locals
```
flash-gordon, line 22: I think, for performance reasons, it makes sense to memoize `scope`. Otherwise, you'll resolve it via an effect every time you access `scope` (or `locals`). For application code, it won't be a problem in 99% of cases, but I would be more cautious in a library. Of course, this is only applicable if `Context` is a class for one-per-render objects

timriley: Unfortunately, while context is indeed a one-instance-per-render object, the `scope` here will change multiple times over the course of that render: for every template or partial, in fact.

See the related code over in `RenderEnvironment` to see this in action:

```ruby
      # TODO:
      # - Give this keyword arguments, it's getting out of control
      # - Rename `scope` to `scope_name` to make it clear it's not the fully built scope
      def template(name, scope, locals, &block)
        template_env = chdir(name)
        scope = self.scope(scope, locals)

        with_render_env(template_env) {
          with_scope(scope) {
            renderer.template(name, scope, &block)
          }
        }
      end

      # TODO: Update this to match how `#template` works and see what the flow-on effects are
      def partial(name, scope, &block)
        with_scope(scope) {
          renderer.partial(name, scope, &block)
        }
      end
```

So unfortunately I don't think we can get away with memoizing here.

Your comment about performance makes me think I should probably look at a building and running a few benchmarks test before merging this code in. Does that seem reasonable to you? Is there anything in particular I should be looking out for, performance-wise?

timriley:
Benchmarks update as of today.

```
Comparison:

render-env-via-reader-effect:    10833.1 i/s
                        main:     8516.4 i/s - 1.27x  (± 0.00) slower
```

timriley:
Closing in favour of #223.

PR #180, opened 2020-10-14
Configure application views with application inflector

timriley:
Configure all views inside an Hanami application with the application's own inflector. This provides consistent inflections inside the view layer.

To do this:

1. Create an `Hanami::View::ApplicationConfiguration`, to provide the application-level `Application.config.views` configuration object, just like we already do for `Application.config.actions`
2. Hide the standard view `inflector` setting so that a separate inflector cannot be configured for views
3. Then pass the application-level inflector instead when setting up any `ApplicationView`

This PR relies upon and will be merged with https://github.com/hanami/hanami/pull/1081.

### DIFF
--- a/lib/hanami/view/application_configuration.rb
+++ b/lib/hanami/view/application_configuration.rb
@@ -1,0 +1,60 @@
+# frozen_string_literal: true
+
+require_relative "../view"
+
+module Hanami
+  class View
+    class ApplicationConfiguration
+      def initialize
+        @base_config = View.config.dup
+
+        configure_defaults
+      end
+
+      # Returns the list of available settings
+      #
+      # @return [Set]
+      #
+      # @since 2.0.0
+      # @api private
+      def settings
+        View.settings - NON_FORWARDABLE_METHODS
+      end
+
+      private
+
+      attr_reader :base_config
+
+      def configure_defaults
+        self.paths = ["web/templates"]
+        self.template_inference_base = "views"
+        self.layout = "application"
+      end
+
+      # An inflector for views is not configurable via `config.views.inflector` on an
+      # `Hanami::Application`. The application-wide inflector is already configurable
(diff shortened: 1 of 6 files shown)


timriley:
[Configure application views with application inflector](https://trello.com/c/jJqElBmm/77-configure-application-views-with-application-inflector)

lib/hanami/view/application_configuration.rb
```diff
@@ -0,0 +1,60 @@
...
+      # `Hanami::Application`. The application-wide inflector is already configurable
+      # there as `config.inflector` and will be used as the default inflector for views.
+      #
+      # A custom inflector may still be provided in an `Hanami::View` subclass, via
+      # `config.inflector=`.
+      NON_FORWARDABLE_METHODS = [:inflector, :inflector=].freeze
+      private_constant :NON_FORWARDABLE_METHODS
+
+      def method_missing(name, *args, &block)
+        return super if NON_FORWARDABLE_METHODS.include?(name)
```
jodosha, line 44: @timriley Probably this `return` is redundant because calling `super` will raise a `NoMethodError` exception, which is what we're looking for.

jodosha (approved):

lib/hanami/view/application_configuration.rb
timriley: @jodosha Do you mean the `super` in the else branch? As far as I see it, we still need this guard clause up top because `base_config` _does_ respond to `inflector` and `inflector=`, but in this case we do _not_ want to forward those methods, because we're hiding them in this context.

Does that make sense?

timriley: Since you've ticked this PR, I'm going to merge this now anyway, but happy to tweak this particular bit after further discussion :)

PR #181, opened 2020-10-19
Configure template on all application views

timriley:
Previously, the `template` setting was only configured on _subclasses_ of an application view, presuming that the initial direct subclass of `Hanami::View` was an abstract “base” view, not to used directly.

This was an overly constricting approach, since there are numerous reasons that an application author may want to inherit directly from `Hanami::View` and use that subclass as a fully renderable view.

This change ensures the `template` setting is configured on _all_ descendants of `Hanami::View`.


timriley:
[Application views should have their template configured in all cases](https://trello.com/c/zz1fs1Ww/91-application-views-should-have-their-template-configured-in-all-cases)

PR #182, opened 2020-10-22
Add parts_path application config setting

timriley:
This accepts a value like “views/parts” and is later converted into a `part_namespace` on application views, with the namespace being looked up relative to the view’s containing slice or application. For example, for a view In the `Main` slice, the resulting `part_namespace` would become `Main::Views::Parts`.


timriley:
[Configure application views with Parts namespace](https://trello.com/c/DezBMazV/93-configure-application-views-with-parts-namespace)

PR #183, opened 2020-12-23
Custom anonymous scope

jodosha:
## Feature

Introduce the ability to create custom anonymous scopes for logic that is specific to a view. **This is syntactic sugar to avoid end users to create concrete scope classes**.

Custom anonymous scopes have access to rendering context, locals, exposures, and parts.
The methods defined by custom anonymous scopes are available in templates.

### Basic usage

```ruby
class MyView < Hanami::View
  # ... other settings
  scope do
    def greeting
      _locals[:message].upcase + "!"
    end
  end
end

MyView.new.call(message: "Hello") # => "HELLO!"
```

The code above is **equivalent** to the following code, **but it avoids to create a concrete scope class**:

```ruby
class MyScope < Hanami::View::Scope
  def greeting
    _locals[:message].upcase + "!"
  end
end

class MyView < Hanami::View
  config.scope = MyScope.new
end
```

### Inherited scope

If the view is inheriting a scope from a super class, or it already has some explicit scope, then the custom anonymous scope will automatically inherit from it.

```ruby
class ApplicationScope < Hanami::View::Scope
  private

  def formatted_timestamp(time)
    time.strftime("...")
  end
end

class ApplicationView < Hanami::View
  config.scope = ApplicationScope
end

class MyView < ApplicationView
  scope do
    def last_updated_at
      formatted_timestamp(_locals[:article].updated_at)
    end
  end
end

MyView.new.call(article: OpenStruct.new(updated_at: Time.now.utc)) # => "2020-12-23 13:39:57"
```

### DIFF
--- a/lib/hanami/view/standalone_view.rb
+++ b/lib/hanami/view/standalone_view.rb
@@ -1,3 +1,5 @@
+require_relative "scope"
+
 module Hanami
   class View
     module StandaloneView
--- a/lib/hanami/view/standalone_view.rb
+++ b/lib/hanami/view/standalone_view.rb
@@ -152,6 +152,89 @@ module Hanami
 
         # @!endgroup
 
+        # @!group Scope
+
+        # Creates and assigns a scope for the current view.
+        #
+        # The newly created scope is useful to add custom logic that is specific
+        # to the view.
+        #
+        # The scope has access to locals, exposures, and inherited scope (if any)
+        #
+        # If the view already has an explicit scope the newly created scope will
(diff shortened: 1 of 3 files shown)


timriley (changes requested):
@jodosha This is wonderful! Thanks for piecing it together 😄 

I think there's just one thing we need here: a `require_relative "scope"` at the top of this file, since we're now referring to the `Scope` class directly in the default argument for the new `.scope` method.

timriley (approved):
Lovely!

PR #185, opened 2021-02-27
Guard on Array of names being provided with singular value 

adam12:
A singular value will trigger the constantization of the names Array,
which returns an invalid result. The user likely wanted to decorate
an Array-like collection if they were manually providing an array of
names through the `as` option.

### DIFF
--- a/lib/hanami/view/part_builder.rb
+++ b/lib/hanami/view/part_builder.rb
@@ -49,6 +49,10 @@ module Hanami
       private
 
       def build_part(name, value, **options)
+        if options[:as].is_a?(Array)
+          raise ArgumentError, "Unable to use Array for :as when exposing not array-like #{value.class}"
+        end
+
         klass = part_class(name: name, **options)
 
         klass.new(
--- a/spec/unit/part_builder_spec.rb
+++ b/spec/unit/part_builder_spec.rb
@@ -163,6 +163,14 @@ RSpec.describe Hanami::View::PartBuilder do
           expect(part.profile).to be_a Test::Parts::Profile
         end
 
+        describe "alternative names provided via :as option" do
+          let(:options) { {as: [:admin_user, :admin_collection]} }
+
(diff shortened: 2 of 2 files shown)


spec/integration/part_builder_spec.rb
```diff
@@ -68,6 +68,51 @@ def each(&block)
...
+
+    it "supports wrapping an array and its members in custom part classes provided to exposure :as option with non-ary block argument" do
+      view = Class.new(Hanami::View) do
+        config.paths = SPEC_ROOT.join("fixtures/templates")
+        config.layout = nil
+        config.template = "decorated_parts"
+        config.part_namespace = Module.new
+
+        expose :customs, as: [Test::CustomArrayPart, Test::CustomPart] do
+          Object.new
```
adam12, line 102: This non-array like object causes `resolve_part_class` to be called with the array provided in `as` and not a single value.

It's only triggered if `config.part_namespace` has been set to a non-nil value.

Should we raise an exception if a value is provided that doesn't respond to `to_ary` but an array of parts has been provided as an option through `as`?

timriley: Yep, I think it'd make sense to raise an ArgumentError at the top of `PartBuilder#build_part` if `options[:as]` is an array. By the time `#build_part` is called, the `:as` option should either be a singular name (or class).

I think by raising the error in this method (as opposed to a deeper method, e.g. `#part_class`) and raising specifically if it's an array, that's our best bet for returning an actually useful error message for the particular scenario you've raised. What do you think?

timriley: Also, I'm not really sure that these additional integration examples using the block form of `expose` are adding much value here, since by the time we get the part building, _where_ the value actually came from (either via an expose block or from being passed directly) doesn't really matter.

Is there a way you can exercise the behaviour you're looking to update without replicating multiple tests?

PR #186, opened 2021-02-27
Fix spelling in specs

adam12:
Just a few spelling fixes.


timriley:
Gosh, I can't believe there so many! 😅 Thanks Adam.

PR #190, opened 2021-05-20
Use setting kwargs to fit new dry-configurable API

timriley:
This updates the `setting` declarations to use the updated dry-configurable API, which uses `default:` and `constructor:` instead of second positional arg and trailing block arg respectively.

This will also unbreak the view integration tests on hanami/hanami for Ruby 2.6 and 2.7, due to their misinterpretation of a hash argument as keyword arguments.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,11 +2,12 @@
 
 source "https://rubygems.org"
 
-git_source(:github) { |repo_name| "https://github.com/#{repo_name}" }
-
 eval_gemfile "Gemfile.devtools"
 
 gemspec
+
+gem "dry-configurable", github: "dry-rb/dry-configurable"
+gem "dry-system", github: "dry-rb/dry-system"
 
 group :tools do
   gem "hotch"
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -8,6 +8,8 @@ eval_gemfile "Gemfile.devtools"
 
 gemspec
 
+gem "dry-configurable", git: "https://github.com/dry-rb/dry-configurable.git", branch: "master"
```
waiting-for-dev, line 11: Hey @timriley , is this new line supposed to be here, or is it a leak from your debugging session? :slightly_smiling_face:

timriley: @waiting-for-dev Thanks for checking :) It's needed here for the time being, to allow the tests on the master branch to pass. Without this they'll fail because they don't work with the latest release of the dry-configurable gem.

When it comes to preparing all these changes for release, I'll do it in the right order (e.g. dry-configurable first, others afterwards) and pull out these Gemfile entries before each release.

PR #191, opened 2021-06-09
Raise LayoutNotFoundError when layout is missing

pvcarrera:
This makes the template errors more granular. Layout errors will have their own error which should provide a better "layout missing" error.

### DIFF
--- a/lib/hanami/view/errors.rb
+++ b/lib/hanami/view/errors.rb
@@ -25,5 +25,20 @@ module Hanami
         super(msg)
       end
     end
+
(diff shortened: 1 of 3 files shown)


adam12 (approved):
Implementation looks good to me. Ran specs locally.

Let's see if @timriley has any thoughts.

timriley:
[Provide a better "layout is missing" exception for (misconfigured) views](https://trello.com/c/NPCXl1Ji/52-provide-a-better-layout-is-missing-exception-for-misconfigured-views)

PR #196, opened 2021-12-02
Make application context aware of its provider

waiting-for-dev:
It allows us to extract information depending on the given provider
(slice vs. application). Plus, it makes it consistent with the application
view and controller.

### DIFF
--- a/lib/hanami/view/application_context.rb
+++ b/lib/hanami/view/application_context.rb
@@ -2,33 +2,52 @@
 
 module Hanami
   class View
-    module ApplicationContext
-      def initialize(inflector: Hanami.application.inflector, **options)
-        @inflector = inflector
-        super
+    class ApplicationContext < Module
+      attr_reader :provider
+      attr_reader :application
+
+      def initialize(provider)
+        @provider = provider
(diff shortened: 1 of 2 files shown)


jodosha:
@timriley I'll leave this to your expertise 🙂

lib/hanami/view/application_context.rb
```diff
@@ -2,33 +2,56 @@
...
+        define_initialize
+        context_class.include(InstanceMethods)
       end
 
-      def session
-        request.session
+      private
+
+      def define_initialize
+        resolve_inflector = method(:resolve_inflector)
```
timriley, line 22:
```suggestion
        inflector = application.inflector
```

Might be worth keeping this simple and only moving to a method if the complexity grows?

timriley (approved):
Looks good, @waiting-for-dev, thank you! Would you mind making that one minor tweak before merging? Happy for you to merge right after :)

lib/hanami/view/application_context.rb
waiting-for-dev: Makes sense, @timriley, done!!

PR #197, opened 2021-12-02
Access application routes from the application context

waiting-for-dev:
Relates to https://github.com/hanami/hanami/pull/1119

### DIFF
--- a/spec/integration/context/application_context/routes_spec.rb
+++ b/spec/integration/context/application_context/routes_spec.rb
@@ -1,0 +1,64 @@
+# frozen_string_literal: true
+
+require "hanami"
+require "hanami/view/context"
+
+RSpec.describe "Application context / Routes", :application_integration do
+  it "accesses application routes" do
+    with_tmp_directory(Dir.mktmpdir) do
+      write "config/application.rb", <<~RUBY
+        require "hanami"
+
+        module TestApp
+          class Application < Hanami::Application
+          end
+        end
+      RUBY
+
+      write "config/routes.rb", <<~RUBY
+        module TestApp
+          class Routes < Hanami::Application::Routes
+            define do
+              slice :main, at: "/" do
+                root to: "test_action"
+              end
+            end
+          end
+        end
+      RUBY
+
+      write "lib/test_app/view/context.rb", <<~RUBY
+        module TestApp
+          module View
+            class Context < Hanami::View::Context
+            end
+          end
+        end
+      RUBY
+
+      require "hanami/init"
+
+      context = TestApp::View::Context.new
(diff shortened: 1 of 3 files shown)


spec/integration/context/application_context/routes_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+            class Context < Hanami::View::Context
+            end
+          end
+        end
+      RUBY
+
+      require "hanami/init"
+
+      context = TestApp::View::Context.new
+      expect(context.routes.path(:root)).to eq "/"
```
waiting-for-dev, line 42: @timriley, for some reason, this test is flaky. It seems that sometimes` Hanami.application[:routes_helper]` is not defined, so `context.routes` ends up being `nil`. Do you have an idea what could be going on?

timriley: @waiting-for-dev Definitely happy to take a look. Do you have an RSpec command and a seed I can use to replicate this, by any chance?

waiting-for-dev: Thanks, @timriley. Sure! Here it goes:

```bash
bundle exec rspec --seed=26933 spec/integration/context/application_context/routes_spec.rb
```

I thought that maybe something would come up to your mind quickly, but if not, I'll also look further into it :slightly_smiling_face:

waiting-for-dev: If you run tests for that single file, the error only happens when the problematic test is run in the second place. Probably, it doesn't work when another integration test has been run before. I'd say that we're missing something in the teardown step. I tried with some tweaks, but no success so far. It seems related to Hanami itself and not Zeitwek. If I rename `TestApp` in one of the two tests, the problem remains.

It's the only place where we're using `with_tmp_directory` if that helps.

timriley: Fixed in https://github.com/hanami/view/pull/197/commits/6f848d8ab175b09ced82e06443c728fb4861180b 😄

timriley:
@waiting-for-dev Fixed that test issue for you! Is this PR otherwise ready to go, in your eyes? (Just noting that it's still marked as draft)

waiting-for-dev:
> @waiting-for-dev Fixed that test issue for you! Is this PR otherwise ready to go, in your eyes? (Just noting that it's still marked as draft)

Thanks, @timriley! :heart: :heart: Yeah, it's ready to go, it was a draft just because the flaky test :slightly_smiling_face:

timriley:
Thanks @waiting-for-dev 😄

PR #200, opened 2022-01-12
Remove examples/ folder

cllns:
They're out of date and, while integrations are important, they belong in documentation and not in the codebase.


timriley:
Thanks @cllns 😄

PR #203, opened 2022-02-04
Views structure

jodosha:
## Enhancements

  * Inject app `settings` and `assets` (if bundled) into `Hanami::View::ApplicationContext`
  * Temporary ported from `hanami/hanami-2-application-template` a few helpers: `#content_for`, `#current_path` `#csrf_token`. Some of those helpers will be moved to `hanami-helpers` gem.

## Fixes

  * Ensure `Hanami::View::ApplicationView` to handle deep inheritance chain (app base view -> slice base view -> slice view) https://github.com/hanami/view/pull/203/commits/a01d0a9e88f5510f1e0a6c181ef82c9c59b344f0 

## Changes

  * Default parts namespace is now `view/parts` (it was `views/parts`)
  * [internal] Changed the integration specs to reflect the new Views structure (see https://github.com/hanami/hanami-2-application-template/pull/76)

---

Ref https://github.com/hanami/hanami/pull/1151
Ref https://github.com/hanami/hanami-2-application-template/pull/76

### DIFF
--- a/lib/hanami/view/application_context.rb
+++ b/lib/hanami/view/application_context.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "hanami/view/errors"
+
 module Hanami
   class View
     class ApplicationContext < Module
@@ -20,11 +22,15 @@ module Hanami
 
       def define_initialize
         inflector = application.inflector
+        settings = application[:settings] if application.key?(:settings)
         routes = application[:routes_helper] if application.key?(:routes_helper)
+        assets = application[:assets] if application.key?(:assets)
 
         define_method :initialize do |**options|
           @inflector = options[:inflector] || inflector
+          @settings = options[:settings] || settings
           @routes = options[:routes] || routes
+          @assets = options[:assets] || assets
           super(**options)
         end
       end
@@ -32,6 +38,36 @@ module Hanami
       module InstanceMethods
         attr_reader :inflector
         attr_reader :routes
+        attr_reader :settings
+
+        def initialize(**args)
+          defaults = {content: {}}
+
+          super(**defaults.merge(args))
+        end
+
+        def content_for(key, value = nil, &block)
+          content = _options[:content]
+          output = nil
+
+          if block
+            content[key] = yield
+          elsif value
+            content[key] = value
+          else
+            output = content[key]
+          end
+
+          output
+        end
+
+        def current_path
+          request.fullpath
+        end
+
+        def csrf_token
+          request.session[Hanami::Action::CSRFProtection::CSRF_TOKEN]
+        end
 
         def request
           _options.fetch(:request)
@@ -43,6 +79,11 @@ module Hanami
 
         def flash
           response.flash
+        end
+
+        def assets
+          @assets or
+            raise Hanami::View::MissingProviderError.new("hanami-assets")
         end
 
         private
--- a/spec/integration/application_view/part_namespace_spec.rb
+++ b/spec/integration/application_view/part_namespace_spec.rb
@@ -20,29 +20,38 @@ RSpec.describe "Application view / Part namespace", :application_integration do
     Hanami.application.register_slice :main, namespace: Main, root: "/path/to/app/slices/main"
 
     Hanami.prepare
+
+    module TestApp
+      module View
+        class Base < Hanami::View
+        end
+      end
+    end
   end
 
   context "view in slice" do
     let(:view_class) {
       module Main
-        class View < Hanami::View
+        module View
+          class Base < TestApp::View::Base
+          end
         end
       end
 
-      Main::View
+      Main::View::Base
     }
 
     context "parts_path configured" do
       let(:application_hook) {
         proc do
-          config.views.parts_path = "views/custom_parts"
+          config.views.parts_path = "view/custom_parts"
         end
       }
 
       context "namespace exists" do
         before do
           module Main
-            module Views
+            module View
               module CustomParts
               end
             end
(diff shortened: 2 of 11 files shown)


cllns:
Duplicate of #199 😉

jodosha:
@cllns Thanks. I didn't know.

BTW, this is a misleading PR title, because I wanted to package all the changes here.
We can merge yours and I'll rebase.

But let's wait a couple of days until we ship alpha6.

spec/integration/application_view/part_namespace_spec.rb
```diff
@@ -95,43 +104,37 @@ module CustomParts
...
+            module View
               module CustomParts
               end
             end
           end
         end
 
-        it "is the matching module within the slice" do
-          is_expected.to eq TestApp::Views::CustomParts
+        # FIXME: @jodosha ask @timriley
```
jodosha, line 126: @timriley This is failing and IDK why. Could you please help me? Thanks 🙏

spec/integration/application_view/part_namespace_spec.rb
```diff
@@ -141,8 +144,9 @@ module CustomParts
           }
         end
 
-        it "is the matching module within the slice" do
-          is_expected.to eq TestApp::Views::CustomParts
+        # FIXME: @jodosha ask @timriley
```
jodosha, line 147: @timriley This is failing and IDK why. Could you please help me? Thanks 🙏

lib/hanami/view/application_context.rb
```diff
@@ -20,18 +22,52 @@ def included(context_class)
...
+
+          output
+        end
+
+        def current_path
+          request.fullpath
+        end
+
+        def csrf_token
+          request.session[Hanami::Action::CSRFProtection::CSRF_TOKEN]
```
timriley, line 69: The fact that this references out to `Hanami::Action` from within hanami-view is definitely a smell to me, and yet another sign that we should just pull all of these out into the hanami gem itself. Looking forward to doing this soon.

In the meantime, I presume we're happy with this constant being referenced without a corresponding `require` given that this is intended only to be used inside a full app with all gems present, yes?

lib/hanami/view/application_context.rb
```diff
@@ -45,6 +81,11 @@ def flash
           response.flash
         end
 
+        def assets
+          @assets or
+            raise Hanami::View::MissingProviderError.new("hanami-assets")
```
timriley, line 86: I'm not fully sold on the name of this error class, but I'm good for us to leave it as is and we can reconsider it once we move all this integration stuff into hanami itself (a better name might present itself there once we're seeing it in the context of all the integrations together).

spec/integration/application_view/part_namespace_spec.rb
timriley: @jodosha This was a load order issue with the tests (the view class was defined in the top-level `before` block, which ran before the context-level `before` block just above, which meant that it couldn't see the `CustomParts` module, since it wasn't defined yet.

Have pushed up a fix :)

spec/integration/application_view/part_namespace_spec.rb
timriley: Sorted per the same fix as above :)

timriley (approved):
Awesome stuff, Luca!

spec/integration/application_view/part_namespace_spec.rb
jodosha: Thank you 🙏

spec/integration/application_view/part_namespace_spec.rb
jodosha: Thank you. 🙏

lib/hanami/view/application_context.rb
jodosha: I assume that views are present only when actions are there too. But not viceversa (views without actions).

lib/hanami/view/application_context.rb
```diff
@@ -20,18 +22,52 @@ def included(context_class)
 
       def define_initialize
         inflector = application.inflector
+        settings = application[:settings] if application.key?(:settings)
         routes = application[:routes_helper] if application.key?(:routes_helper)
+        assets = application[:assets] if application.key?(:assets)
```
solnic, line 27: Unrelated to the PR but: I think we want to change dry-container so that it doesn't raise when doing `#[]`. It's incompatible with `Hash#[]` and as much as I like strictness, I believe it was not a good idea to have behavior like that. We have such `key?` checks in *many places* across the ecosystem which is clearly an indicator that we want a Hash-like behavior instead. It would be great to have dry-container 1.0 released before H2 final, hence I'm bringing this up /cc @timriley

timriley: @solnic I feel like the current behaviour is actually right, in that it should be a hard error if you're trying to fetch something from the container that doesn't exist.

With a container typically serving the purpose of vending all the known components of an application or subsystem of some kind, I think that the consequence of resolving an unknown component should be much like the consequence of trying to refer to an unknown class constant — that is to say, an exception. If the container was much looser in this behavior, it would more likely lead to unexpected errors around `nil` objects further down the line, which isn't confidence-building.

It's worth noting that the `key?` checks you're bringing up are typically inside framework/library/integration-style code, rather than in "userland."  I'm definitely willing to trade off the need to check for a key first before resolving in those cases if we can retain the benefits of stricter resolution for the end users of the container.

As for this case in particular, what we're dealing with is a container with _unknowable contents_, which is why we need these checks. The fact that they add a bit of noise here I don't actually mind, because it makes it clearer that this might not even be a registered component in the first place (e.g. kind of like the extra "noise" you might have when you're unwrapping a Maybe monad).

PR #208, opened 2022-04-22
Ensure proper access to full `locals` hash in templates

timriley:
This _should_ already have worked, thanks to the `method_missing`-based `#locals` method that we expose in `Hanami::View::Scope`.

However, `locals` in in fact a _local variable_ in the template rendering context, since that's the name of the one and only parameter for the method that Tilt compiles each template into. This means we can't access the `#locals` on our `Scope` instance without explicitly calling `self.locals`.

To fix this, we pass our `scope._locals` additionally into Tilt's `render` method, which will allow bare `locals` references to work inside templates.

Since this behavior evaded detection for so many years, we now have a dedicated test to make sure it remains working like we expect in the future.

### DIFF
--- a/lib/hanami/view/renderer.rb
+++ b/lib/hanami/view/renderer.rb
@@ -46,7 +46,7 @@ module Hanami
       end
 
       def render(path, scope, &block)
-        tilt(path).render(scope, &block)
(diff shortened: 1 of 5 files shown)


jodosha (approved):
@timriley LGTM 👍

PR #218, opened 2022-12-01
Include format in MissingTemplateError

radar:
This will now show an error such as:

```
Puma caught this error: Template +books/show+ for format +html+ could not be found in paths:

 - /Users/ryan.bigg/code/playground/bookshelf/app/templates
```

Fixes https://trello.com/c/QqyDKhjq/177-hanami-view-template-not-found-errors-should-include-what-format-it-is-using-to-look-for-them

### DIFF
--- a/lib/hanami/view/errors.rb
+++ b/lib/hanami/view/errors.rb
@@ -23,7 +23,7 @@ module Hanami
     class TemplateNotFoundError < StandardError
       def initialize(template_name, format, lookup_paths)
         msg = [
-          "Template +#{template_name}+ for format +#{format}+ could not be found in paths:",
+          "Template `#{template_name}' for format `#{format}' could not be found in paths:",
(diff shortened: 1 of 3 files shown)


lib/hanami/view/errors.rb
```diff
@@ -21,9 +21,9 @@ def initialize(key)
     #
     # @api private
     class TemplateNotFoundError < StandardError
-      def initialize(template_name, lookup_paths)
+      def initialize(template_name, format, lookup_paths)
         msg = [
-          "Template +#{template_name}+ could not be found in paths:",
+          "Template +#{template_name}+ for format +#{format}+ could not be found in paths:",
```
jodosha, line 26:
```suggestion
          "Template `#{template_name}' for format `#{format}' could not be found in paths:",
```

jodosha (approved):
@radar Thank you 🙏

PR #226, opened 2023-03-06
Introduce new ERB engine, ensure consistent automatic HTML escaping

timriley:
Introduce `Hanami::View::ERB::Template`, a new ERB implementation based on Temple. The key features of this implementation are:

- Auto-escaping any non-`html_safe?` values given to `<%=` ERB expression tags, with auto-escaping disabled when using the `<%==` alternative syntax.
- Implicitly capturing and correctly outputting block content without the need for special helpers. This allows helpers like `<%= form_for(:post) do %>` to be used, with the `form_for` helper itself doing nothing more special than a `yield`.

This exact combination of features is required for the experience we offer across all HTML template engines in hanami-view, but does not exist in any of the existing ERB implementations:

- [erbse](https://github.com/apotonick/erbse), which we used before, supports implicit block capture but not auto-escaping. It is also an incomplete ERB implementation, and otherwise not maintained.
- [erubi](https://github.com/jeremyevans/erubi) supports Rails-specific auto-escaping in its [Rails-specific extension](https://github.com/rails/rails/blob/main/actionview/lib/action_view/template/handlers/erb/erubi.rb), but not implicit block capture).

So to solve this, here we've created our own implementation! It ships with hanami-view so basic ERB templating will now be available without requiring another gem. This new implementation is based on the [Temple](https://github.com/judofyr/temple) template framework, which is now used by both Slim and Haml. In fact, we use the very same [Escapable](https://github.com/judofyr/temple/blob/f8a069bdba5d2d821b1a8c8221b9e6e225e9e1ed/lib/temple/filters/escapable.rb) filter as they do, which means we now have absolute code-level consistency in terms of HTML escaping across the three template engines.

Along with this, there are a few other changes here to provide this HTML-escaping consistency:

- Introduce `Hanami::HTML::SafeString`, which is returned when calling `String#html_safe`. Require this at the top of `lib/hanami/view.rb` to ensure that the `String#html_safe` method is available at all times when using hanami-view. Yes, this is a monkey-patch, but given its existing wide support and low-friction user experience, this feels like the best option for us to take. (We originally used Temple's own SafeString by thanks to @byroot's feedback we've instead created a more complete implementation here instead)
- Remove the "Tilt adapters" that were checked before every render and told people if they had to install certain gems. Instead, register default engines in our default Tilt mapping. This is a much more "confident" posture and is one we can readily take now that we can ship an internal ERB implementation.
- Add a new integration test exercising the auto-escaping for each of the engines.

This PR can be reviewed by commit. The first commit adds the new ERB implementation, while the second addresses the last two dot points above. UPDATE: there are a few additional commits now as I responded to feedback, but they should all be easy enough to follow.


timriley:
[Ensure consistent HTML escaping](https://trello.com/c/ZTIY46ik/21-ensure-consistent-html-escaping)

lib/hanami/view/erb/filters/block.rb
```diff
@@ -0,0 +1,44 @@
...
+        # @since 2.0.0
+        # @api private
+        class Block < Temple::Filter
+          END_LINE_RE = /\bend\b/
+
+          def on_erb_block(escape, code, content)
+            tmp = unique_name
+
+            # Remove the last `end` :code sexp, since this is technically "outside" the block
+            # contents, which we want to cpature separately below. This `end` will be added back
```
radar, line 21:
```suggestion
            # contents, which we want to capture separately below. This `end` will be added back
```

lib/hanami/view/html.rb
```diff
@@ -0,0 +1,83 @@
...
+          self
+        end
+
+        # @return [self]
+        #
+        # @api public
+        # @since 2.0.0
+        def to_s
+          self
+        end
```
byroot, line 46:
```suggestion
        end


        private

        def initialize_copy(other)
          super
          freeze
        end
```

I forgot to point you need to freeze copies as well, otherwise:

```ruby
str = "foo".html_safe.dup
p str.html_safe? # => true
str << "<script>"
p str # => "foo<script>"

```

byroot: Arf, even that won't be enough, many string methods will return you a `SafeString`:

```ruby
str = "foo".html_safe
p str.sub("foo", "<script>").class # Hanami::View::HTML::SafeString
```

So delegation is likely the way to go here. e.g. `class SafeString < DelegateClass(String)`

timriley: @byroot Thanks for circling back here! Great tip about the initialize_copy. I've just added this in https://github.com/hanami/view/pull/226/commits/4647c40125ffd9dcbcaf95bf21a992e15dc54966.

As for your `str.sub` example, I actually tested that during my earlier work here, and even included a test case:

```ruby
  it "returns non-safe strings from methods returning new strings" do
    new_string = safe_string.gsub(/hello/, "world")
    expect(new_string).to eq "world"
    expect(new_string).not_to be_html_safe
  end
```

The test passed so I presumed things to be fine...

If I check things out more manually, here's what I'm seeing:

```ruby
safe_str = "Hello".html_safe
# => "Hello"
safe_str.html_safe?
# => true
safe_str.class
# => Hanami::View::HTML::SafeString

new_str = safe_str.gsub("l", "y")
# => "Heyyo"
new_str.html_safe?
# => false
new_str.class
# => String
```

This is consistent with what the test was exercising... but maybe I'm missing something? Is there another category of string methods that when called on a subclass return a new instance of that subclass rather than a plain `String` itself?

byroot: > Is there another category of string methods that when called on a subclass return a new instance of that subclass rather than a plain String itself?

Nah, my bad, I was testing on 2.7, the behavior of Ruby this changed in 3.0. Hamani being 3.0+ you're good.

Sorry for the false alarm.

timriley: > I was testing on 2.7, the behavior of Ruby this changed in 3.0.

TIL! And no worries at all. I appreciate all your input here! ❤️

byroot: For completeness:

  - https://bugs.ruby-lang.org/issues/10845
  - https://github.com/ruby/ruby/pull/3701

It was indeed changed for all String methods except two, but that are safe in this context (`@+` and `@-`).

lib/hanami/view/erb/engine.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+require "temple"
+require_relative "parser"
+require_relative "filters/block"
```
jodosha, line 5: @timriley Nitpick: please move these `require_relative` inside the module definition.

Ruby Reflection is more straightforward, so when we do `Hanami::View.const_location(:ERB)`, we get this file instead of `filters/block.rb`.

lib/hanami/view/html.rb
```diff
@@ -0,0 +1,104 @@
...
+        # @since 2.0.0
+        def html_safe
+          Hanami::View::HTML::SafeString.new(self)
+        end
+      end
+    end
+  end
+end
+
+class String
```
jodosha, line 73: @timriley Do we have a chance not to do this?

In the context where `#html_safe?` is invoked, can we hook it into a wrapper?

Naive code example:

```ruby
class Wrapper
  def call(obj)
    case obj
    when String
        # ...
    when Numeric
       # ....
    end
  end
end

# This is a fictional object from Temple/Tilt where the escaping happens
class TempleRenderingContext
  def call(obj)
     wrapper.call(obj)
  end
end
```

jodosha (approved):
@timriley Monumental work. Thank you for what you're doing for Hanami. You have my deepest and true gratitude. 💯 🙇

lib/hanami/view/erb/engine.rb
timriley: Done in 6eb4637

> Ruby Reflection is more straightforward, so when we do Hanami::View.const_location(:ERB), we get this file instead of filters/block.rb.

Thanks for sharing this particular detail; I've never thought of it along these lines.

lib/hanami/view/html.rb
timriley: @jodosha Fair question, thank you for asking!

I'm not sure if we have much choice here, because we don't really control every instance of “the context where `#html_safe?` is invoked” — for the three template engines we support (Hanami::View::ERB, Slim, Haml) this chiefly happens [inside Temple's Escapable filter](https://github.com/judofyr/temple/blob/f8a069bdba5d2d821b1a8c8221b9e6e225e9e1ed/lib/temple/filters/escapable.rb#L18).

For Hanami::View::ERB, we swap this filter for one that we fully own, and which uses an approach like you describe, but that still wouldn't help us in terms of providing seamless support for Haml and Slim. They both use the Temple Escapable filter directly, and Haml also has [an additional check](https://github.com/haml/haml/blob/d0a86c3eba4563073e950f9e7307a947d4379a64/lib/haml/util.rb#L31) in its escaping utils. To change this would require us to maintain patches of Haml and Slim that are delicate and prone to breakage.

I'm sure it'd be possible for us to come up with a whole new strategy and slowly introduce all the required pieces into Temple, Haml, Slim and the like, but that feels like a lot of time-consuming work simply for the sake of avoiding this one String/Object/Numeric extension that is already endemic across the Ruby web app ecosystem.

In addition, your suggestion here doesn't cover the case where user-authored helpers might need to return strings flagged as html_safe. In this case, if we decide not to use the `String#html_safe` extension, then users would need to construct their own SafeString instances directly. We could probably provide a helper available in _most_ contexts to do this, e.g. `html_safe(my_str)`, but it would only cover those helpers working in execution contexts that the Hanami framework expressly controls. In other contexts, our users would have to remember to `Hanami::View::HTML::SafeString.new(my_str)` instead. This approach doesn't feel great to me.

So my opinion here is that, on balance, this is one place where we're better off simply following the already-established and widespread convention. Folks are going to be learning a whole bunch of stuff already when they adopt Hanami. Here we can at least take one thing off that list :)

timriley: I'm going to go ahead and merge this PR without any changes here, but I do want to make it clear that I'm definitely open to adopting a different approach if it match the usability and robustness of the current one.

(I won't be pursuing that personally because, as I noted above, I'm pretty happy with the current arrangement, and there's a bunch of other much more pressing things to get done to complete our view layer for 2.0)

timriley:
Thank you for the lovely words, @jodosha! ❤️ 

I'm going to merge this now. Just wanted to give you a heads up about a few adjustments I've made today:

- We no longer need to use `<%== yield %>` (the "no escaping" form of the tag) inside your layout. I moved the location of our `.html_safe` call into `Renderer`, so that _every_ rendering (either a partial or the one rendering of whole template) is considered html_safe
- I updated the slim tests to use `=` (output with auto-escaping) instead of `==` (output without escaping) where possible, and in doing so, discovered and fixed an issue where Slim's auto-escaping wouldn't be enabled if it is required before hanami-view.

See the last few commits if you're curious.

Anyway, big milestone here for hanani-view. Onto the next!

lib/hanami/view/html.rb
jodosha: @timriley Thanks for the extensive explanation. At this point, it’s the most pragmatic path to take.

We’ll reuse your words in our guides as a note for this monkey-patching exception.

PR #227, opened 2023-04-01
Add configurable part_class and scope_class settings

timriley:
Add `part_class` and `scope_class` settings, used by the `part_builder` and `scope_builder` respectively as the default class to use when no value-specific part or scope classes can be found.

These settings default to `Hanami::View::Part` and `Hanami::View::Scope` respectively, so there is no change to hanami-view's behaviour when not explicitly configured.

The reason we're adding these settings, however, is so that full Hanami apps may configured their own custom `part_class` and `scope_class`, providing classes that have mixed in a range of helper modules.

### DIFF
--- a/lib/hanami/view.rb
+++ b/lib/hanami/view.rb
@@ -10,6 +10,7 @@ require_relative "view/context"
 require_relative "view/exposures"
 require_relative "view/errors"
 require_relative "view/html"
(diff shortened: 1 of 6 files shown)


solnic (approved):
LGTM 👏🏻

PR #228, opened 2023-04-03
Add helpers

timriley:
These are taken from hanami-helpers and hanami-utils and simplified and repackaged here in hanami-view.

Given these helpers are largely standalone, and all similar in their general concern, it makes sense for these helpers to be located inside hanami-view, since they would be helpful even in standalone usage of the gem.

_This should be merged after https://github.com/hanami/view/pull/227, since it currently builds on top of it (to allow for easy testing of these combined PRs from hanami/hanami)._


timriley:
[Finish porting rest of helpers 1.x](https://trello.com/c/CmNsXScw/6-finish-porting-rest-of-helpers-1x)

PR #230, opened 2023-04-13
Mark blocks captured from templates as HTML safe

timriley:
This is important for Part or Scope methods (including the helpers we’ll soon be shipping) that want to determine whether to escape strings captured from template blocks based on whether they are `.html_safe?`.

It's the primary job of templates to create HTML, so it makes sense for all template-captured blocks to be marked as HTML safe in the first place.

### DIFF
--- a/lib/hanami/view/html_safe_string_buffer.rb
+++ b/lib/hanami/view/html_safe_string_buffer.rb
@@ -38,6 +38,17 @@ module Hanami
     # @api private
     # @since 2.0.0
     class HTMLSafeStringBuffer < Temple::Generators::StringBuffer
+      # Replace `Temple::Generator::ArrayBuffer#call` (which is used via the superclass of
+      # `StringBuffer`) with the standard implementation from the base `Temple::Generator`.
+      #
+      # This avoids certain specialisations in `ArrayBuffer#call` that prevent `#return_buffer` from
+      # being called. For our needs, `#return_buffer` must be called at all times in order to ensure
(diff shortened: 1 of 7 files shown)


jodosha:
@timriley Bingo! This was one of the problems I ran into during my tests, and I forgot about it.

lib/hanami/view/html_safe_string_buffer.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+
+require "temple"
+
+# Add temple patch from https://github.com/judofyr/temple/pull/113 (which was since rolled back) to
+# demonstrate the need for the `:capture_generator` setting value to be propogated in order to
+# consistently handle nested captures.
+module Temple
```
jodosha, line 8: @timriley Is this still needed, even after https://github.com/judofyr/temple/pull/144 was merged?

jodosha (approved):

lib/hanami/view/html_safe_string_buffer.rb
timriley: Nope, this is no longer needed if someone is using the master branch of Temple.

I'm going to keep this patch here so people can test hanami-view even on released versions of temple, but I'll finesse it a bit to remove the method redefinition warning and update the comment to reflect the now-merged PR.

I'll make sure we revisit this and remove the patch before release: https://trello.com/c/bRyiPQEj/38-remove-templegenerator-patch

PR #233, opened 2023-05-07
Use Zeitwerk for code loading

timriley:


timriley:
[Port hanami-view to Zeitwerk](https://trello.com/c/MfaM7nfb/34-port-hanami-view-to-zeitwerk)

PR #237, opened 2023-09-06
Allow layout to be passed in `View#call`

parndt:
In order to be able to render a view but with a different layout to the
one that is expected, we need to allow the layout to be specified, or to
be removed by passing `nil` or `false`, in the `call` interface of the
`View` class.

With these changes, this now works from a `hanami-controller` action:

```ruby
response.render(view, layout: nil)
```

As does this:

```ruby
view.call(layout: nil)
```

### DIFF
--- a/lib/hanami/view.rb
+++ b/lib/hanami/view.rb
@@ -477,8 +477,8 @@ module Hanami
     # @!endgroup
 
     # @api private
-    def self.layout_path
-      File.join(*[config.layouts_dir, config.layout].compact)
+    def self.layout_path(layout)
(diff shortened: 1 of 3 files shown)


lib/hanami/view.rb
```diff
@@ -530,7 +530,7 @@ def call(format: config.default_format, context: config.default_context, layout:
       locals = locals(rendering, input)
       output = rendering.template(config.template, rendering.scope(config.scope, locals))
 
-      if !!layout
+      if layout
```
parndt, line 533: excellent, I'm glad!

PR #238, opened 2023-09-19
Make `Hanami::View::Rendered` more specs friendly

jodosha:
Ref https://github.com/hanami/rspec/pull/15

### DIFF
--- a/lib/hanami/view/rendered.rb
+++ b/lib/hanami/view/rendered.rb
@@ -50,6 +50,31 @@ module Hanami
         output
       end
       alias_method :to_str, :to_s
+
+      # Matches given input with the rendered view
+      #
+      # @param matcher [String, Regexp] matcher
+      #
+      # @return [TrueClass,FalseClass]
+      #
+      # @api public
+      # @since 2.1.0
+      def match?(matcher)
+        output.match?(matcher)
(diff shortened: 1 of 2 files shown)


timriley (approved):
This looks good to me, @jodosha :)

Another option here would be to turn `Rendered` into a subclass of `DelegateClass(String)` so that it exposes the full `String` interface while allowing us to provide access to the locals.

We currently do that using `Rendered#[]`, but if we made it a full delegate class of String, I think we should just expose `#locals` and let people use that directly. On reflection, that's probably a much less surprising interface to users.

What do you think?

In the meantime, I'd be happy to merge this and then tee up a follow-up task to switch to this more string-like alternative.

jodosha:
@timriley Nope. Old, but Gold: https://tenderlovemaking.com/2014/06/02/yagni-methods-are-killing-me.html

timriley:
@jodosha Fair enough, though in a parallel universe, the actual return value of `View#call` would be literal `String` and no one would blink :) This is why I felt it might be a reasonable change here. But yes, keeping this as `Rendered` with a `#to_str` and a small handful of the most useful methods is a much more defensible position, so I'm happy to keep it as is 👍🏼

PR #247, opened 2023-11-04
Remove LayoutNotFoundError and raise TemplateNotFoundError in all cases

### LINKED ISSUE #246: Missing templates result in (incorrect) missing layout error
Hey folks, just giving the migration to 2.1.0.rc1 a shot, and have come across a misleading error - something that cropped up when I last poked around with the earlier beta releases too:

If a template file doesn't exist, the error that I get is that the _layout_ can't be found.

https://github.com/hanami/view/blob/89fa27842d96074ef9d912c99caf6e822229f190/lib/hanami/view.rb#L528-L543

If I comment out the rescue and the line below it - lines 540 & 541 - then I get the original TemplateNotFoundError bubbling up instead, and that is actually useful because it tells me the missing template. I presume that rescue's in there for a reason though, so I'm not sure what the appropriate fix should be!

(FWIW: Ruby 3.2.2, hanami-view 2.1.0.rc1, macOS Ventura)

timriley:
Remove the specialized LayoutNotFoundError, because this was raising incorrect error if there were _missing partials_ being rendered inside the layout.

It is better that we raise a genuine, non-misleading error in these cases. The LayoutNotFoundError only provided marginal utility anyway, and the inclusion of the “layouts/“ prefix for a TemplateNotFoundError raised for a missing layout will still make it clear where the template was being looked for.

Fixes #246.


timriley:
The LayoutNotFoundError was introduced in #191 and has never been formally released, so it is safe to remove here.

spec/unit/view_spec.rb
```diff
@@ -64,8 +64,30 @@ def title
...
+
+  describe "template rendering" do
+    it "raises a TemplateNotFoundError when the template cannot be found" do
+      view = Class.new(Hanami::View) {
+        config.paths = SPEC_ROOT.join("fixtures/templates")
+        config.layout = nil
+        config.template = "missing_template"
+      }.new
+
+      expect { view.() }.to raise_error Hanami::View::TemplateNotFoundError, /Template `missing_template' for format `html' could not be found/
```
jodosha, line 80: @timriley In case of Hanami app, will the real error message appear with the relative path of the template?

Example:

```
Template `app/templates/layouts/foo.html.erb' for format `html' could not be found
```

As a dev, I can just copy and paste the path and create the missing template.

jodosha (commented):
@timriley I agree with the change. Please clarify my inline question.

spec/unit/view_spec.rb
timriley: @jodosha Here's the exact error:

```
     Hanami::View::TemplateNotFoundError:
       Template `layouts/app' for format `html' could not be found in paths:

        - /Users/tim/Source/scratch/bookshelf/app/templates
```

So yes, you could join `layouts/app` onto `/Users/tim/Source/scratch/bookshelf/app/templates` and there you have the correct path to use (minus the file extension).

We can't put an _exact_ file extension on this because we do support fuzzy loading e.g. from any `*.html.*` template files.

I think we _could_ probably finesse this error message to make it even nicer. For example, if there's only one path configured, then we could make the error a lot closer to your example.

However, I'd like to leave that as an exercise for a future enhancement rather than block on this now. This would be a great first issue for a potential Hanami contributor, for example 😄

jodosha (approved):

Issue #252, opened 2023-12-04
Bug: Incorrectly resolved scopes due to the caching


swilgosz:
### Background

When you have two scopes used, in the single view rendering, and you'll use `path` to the scope file instead of a class, you'll get only the first scope loaded, because of the caching.

Checking the code of the scope builder shows, that it allows to properly recognize the scope by file path, but because of caching, it falls back to the last defined.

![image](https://github.com/hanami/view/assets/8088317/6fd5e396-7247-4ea5-8211-3fcd4bb9b4cb)

### To Reproduce

TODO: Link the sample App.

I have the app having 2 scopes in use:

```ruby
 scope(
    "shared/forms/input",
    f: f,
    field_name: :username,
    errors: errors(:username)
      ).render("shared/forms/text_field")
```

```ruby
scope('shared/flash', flash: _context.flash).render('layouts/flash_msg')
```

When I use both used on the same page document, it appears, that the code is tried to be loaded with the first loaded scope class.
`Flash Message#visible`? - is called on the `forms/input` scope definition, because:

1. Flash is called in the layout (resolved later)
2. the `form_input` is used in the view partial, which is resolved first, to be injected into the layout

![image (1)](https://github.com/hanami/view/assets/8088317/b3b50d8e-9fe4-406a-8d29-f7334a2942ac)

**NOTE: ** This is not a case if you pass ScopeClass as an argument.

QuickFix: https://github.com/hanamimastery/app/commit/1dadcbf5e4be0227508107190ed9e0e02beb4ada

**Expected behavior**

I expect the scope builder to properly resolve scopes, no matter if I pass class name, or the file path as argument, and not matter about the ordering of execution.

**Your environment**

- Affects my production application: **YES**
- Ruby version: 3.2.2
- OS: All

PR #253, opened 2023-12-07
Fix: Add support for rendering multiple scopes by passing string as names

swilgosz:
## Background

Resolves: #252 

### Bug 1. 

When scope name is passed as string (either scope class name, or the path to the file), the name resolving goes through the `cache`, with a hardcoded key `scope_class`, not allowing mutliple scopes to be stored and fetched.

### Bug 2.

If namespace is not defined (nil), `const_defined?` is called on `nil`, causing Undefined method error to be risen.

## Design

- Add spec covering both issues detected.
- Change the hardcoded cache key to the one taking passed name into the consideration
- Add support to `nil` namespace, falling back to the Object, so we can use `const_get` and `const_defined?` methods.

### DIFF
--- a/lib/hanami/view/scope_builder.rb
+++ b/lib/hanami/view/scope_builder.rb
@@ -31,7 +31,7 @@ module Hanami
           elsif name.is_a?(Class)
             name
           else
-            View.cache.fetch_or_store(:scope_class, rendering.config) do
+            View.cache.fetch_or_store("scope_class: #{name}", rendering.config) do
               resolve_scope_class(name: name, rendering: rendering)
             end
           end
@@ -40,7 +40,7 @@ module Hanami
         def resolve_scope_class(name:, rendering:)
           name = rendering.inflector.camelize(name.to_s)
 
-          namespace = rendering.config.scope_namespace
+          namespace = rendering.config.scope_namespace || Object
 
           # Give autoloaders a chance to act
           begin
--- a/lib/hanami/view/scope_builder.rb
+++ b/lib/hanami/view/scope_builder.rb
@@ -40,7 +40,7 @@ module Hanami
         def resolve_scope_class(name:, rendering:)
           name = rendering.inflector.camelize(name.to_s)
 
-          namespace = rendering.config.scope_namespace || Object
+          namespace = rendering.config.scope_namespace
 
           # Give autoloaders a chance to act
           begin
--- a/lib/hanami/view/scope_builder.rb
+++ b/lib/hanami/view/scope_builder.rb
@@ -31,7 +31,7 @@ module Hanami
           elsif name.is_a?(Class)
             name
           else
-            View.cache.fetch_or_store("scope_class: #{name}", rendering.config) do
+            View.cache.fetch_or_store(name, rendering.config) do
               resolve_scope_class(name: name, rendering: rendering)
             end
           end
--- a/spec/integration/scope_builder_spec.rb
+++ b/spec/integration/scope_builder_spec.rb
@@ -27,5 +27,31 @@ RSpec.describe "scope builder" do
 
       expect(scope).to be_an_instance_of scope_class
     end
+
+    context 'when multiple scopes are rendered in the same view' do
+
+      let(:builder) { described_class.new }
+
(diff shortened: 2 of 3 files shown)


bkuhlmann:
I ran into this too. Without this patch, use of multiple scopes is essentially useless which defeats the power they provide. It would be good to get this folded in and a patch released shortly after.

lib/hanami/view/scope_builder.rb
```diff
@@ -31,7 +31,7 @@ def scope_class(name = nil, rendering:)
           elsif name.is_a?(Class)
             name
           else
-            View.cache.fetch_or_store(:scope_class, rendering.config) do
+            View.cache.fetch_or_store("scope_class: #{name}", rendering.config) do
```
bkuhlmann, line 34: :bulb: If it helps, in my local patch, I only use `name` which is all you need to store in the cache. Example:

``` ruby
View.cache.fetch_or_store name, rendering.config do
  # Implementation details.
end
```

spec/integration/scope_builder_spec.rb
```diff
@@ -27,5 +27,31 @@
 
       expect(scope).to be_an_instance_of scope_class
     end
+
```
bkuhlmann, line 30: :bulb: If it helps, in my local patch, I needed a way to test that the cache works as expected and ended up with the following spec:

<details>
  <summary>Spec Snippet</summary>
  
``` ruby
RSpec.describe Milestoner::Views::ScopeBuilder do
  subject(:builder) { described_class }

  let :view do
    Class.new Hanami::View do
      config.paths = Bundler.root.join "lib/milestoner/templates"
      config.template = "n/a"
    end
  end

  let(:scope_a) { Class.new Hanami::View::Scope }
  let(:scope_b) { Class.new Hanami::View::Scope }

  before do
    stub_const "TestScopeA", scope_a
    stub_const "TestScopeB", scope_b
    Hanami::View.cache.cache.clear
  end

  describe ".call" do
    it "caches scope without duplicates" do
      builder.call TestScopeA, locals: {}, rendering: view.new.rendering
      builder.call TestScopeB, locals: {}, rendering: view.new.rendering
      builder.call "TestScopeA", locals: {}, rendering: view.new.rendering
      builder.call "TestScopeB", locals: {}, rendering: view.new.rendering
      builder.call scope_a, locals: {}, rendering: view.new.rendering
      builder.call scope_b, locals: {}, rendering: view.new.rendering

      expect(Hanami::View.cache.cache.values).to eq([TestScopeA, TestScopeB])
    end
  end
end
```
</details>

Might be a good idea to add a spec for `#call` because that ensures the cache is used and doesn't duplicate itself. This also means you'd need to clear the cache between specs which why I put the in the `before` block.

bkuhlmann:
Hey @cllns :wave: Any chance this could be folded in as this is becoming a very big issue because without this fix, multiple scopes are effectively useless in Hanami. I've been using this patch for quite some time and haven't hit issues as of yet. I also use this heavily in [Milestoner](https://alchemists.io/projects/milestoner). Here's the full implementation (plus spec) which is nearly identical to the above:

- [Implementation](https://github.com/bkuhlmann/milestoner/blob/main/lib/milestoner/views/scope_builder.rb)
- [Spec](https://github.com/bkuhlmann/milestoner/blob/main/spec/lib/milestoner/views/scope_builder_spec.rb)

timriley:
@bkuhlmann Sorry for the delay. I'll check this one out today. Please stand by.

timriley:
Thanks for doing this @swilgosz! And I'm really sorry for the delay in handling it.

I've merged this now and will release it shortly.

Before merging this, I:

- Expanded and tweaked the tests a bit
- Reduced the scope of this change. Stop falling back to `Object` for scope class lookups if `config.scope_namespace` is not defined. This is not the design intention here; custom scopes should always come from a namespace, so we expect a `scope_namespace` to be configured. I've filed #264 so we can improve the error messaging in this situation.

timriley:
Thanks also for all the helpful info here, @bkuhlmann 😄

timriley:
Have released this as v2.2.1!

PR #274, opened 2026-03-16
Default to undecorated exposures

timriley:
Change `expose` so that it no longer decorates by default, and ask the user to indicate which exposures should be decorated:

- `expose :foo` is now the equivalent of `expose :foo, decorate: false`
- `decorate :foo` is a new class-level method, a convenience that equates to `expose :foo, decorate: true`

This addresses a couple of downsides to our previous behaviour of decorating every exposure by default:

- Users may be surprised to find their values wrapped in parts
- There is a performance impact of wrapping every object in a part; this may be felt when exposing large collections from the view

To restore decorated exposures by default, set `config.decorate_exposures = true` in your view class.

### DIFF
--- a/lib/hanami/view.rb
+++ b/lib/hanami/view.rb
@@ -235,6 +235,23 @@ module Hanami
     # @!scope class
     setting :inflector, default: Dry::Inflector.new
 
+    # @overload config.decorate_exposures=(value)
+    #   Controls whether exposures are decorated by default.
+    #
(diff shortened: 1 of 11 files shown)


parndt (approved):
I'm a big fan of this change

fz-ryanbigg (approved):

radar (approved):

timriley:
Thanks for checking this out, folks :)

PR #275, opened 2026-04-22
Reference Hanami Action instead of Controller

katafrakt:
The gem has been renamed to hanami-action, so the Gemfile needs fixing. This also change other repo names to new canonical form (without leveraging Github redirects)


timriley:
Thanks!

PR #276, opened 2026-05-05
Performance: Cache config

cllns:
~28% fewer allocations (138 -> 99) per render
~18% fewer bytes used (8,739 -> 7,160) per render

+36.8% iteration speed (30,891 ips -> 42,266 ips, 32.4μs -> 23.7μs per render)

(Updated these numbers. The early improvements weren't as high but I closed some apps 😇 and these seem to be consistent)

Using benchmarks/view_benchmark.rb on Ruby 4 (on an M1 Pro).

I originally had this with a class-level `self.cached_config` which isn't necessary so I removed it. This means each instantiation of a view has to compute `cached_config` for itself, adding some overhead. Since we've built this library based on the assumption that a single View will be re-used (rather than created new each time it's used), I think that's fine. Open to other perspectives though!

Another option is using `reader: true` for all of the settings. I experimented with this too, but it's not as clean, because the settings are set on the `View` class rather than the config, so `view_class` has to be passed around everywhere, which hides the intent of sending just config settings around. It's got basically the same performance as this approach though.


timriley:
Thanks for looking into this stuff, @cllns! My question is whether we might actually be able to solve this at a Dry Configurable level, rather than create a workaround for it here? We use Dry Configurable in many other places, so a solve there once would benefit us more.

cllns:
@timriley I was already working on it :) Just had to figure out the right API. See: https://github.com/dry-rb/dry-configurable/pull/167 and updated changes here. 

I made a few tiny refactors to move things into private visibility. There's more I want to do, but I'll do that as a follow-on, so this PR doesn't get too complicated.

cllns:
Switched the Gemfile to use dry-configurable from 'main' since https://github.com/dry-rb/dry-configurable/pull/167 has been merged

lib/hanami/view/renderer.rb
```diff
@@ -55,10 +51,12 @@ def partial(name, format, scope, &block)
 
       private
 
+      attr_reader :config_data, :prefixes
```
timriley, line 47: Just a note on why these were public before: I've never really cared about leaving public attr readers on objects, even if they're not formally "public API". It's Ruby, so if anyone _really_ wanted to get to some object, there's always a way they can do it, so I never really tried to fight it.

This is not to say I'm opposed to this change, but I thought I'd just share some of the rationale I held when putting these things together in the first place :)

Happy for this change to go in as part of this PR, anyway :)

lib/hanami/view/rendering.rb
```diff
@@ -7,28 +7,43 @@ class View
...
 
       # @api private
       # @since 2.1.0
       attr_reader :inflector, :part_builder, :scope_builder
 
+      # @api private
+      # @since 2.3.0
+      attr_reader :part_class, :part_namespace, :scope_class, :scope_namespace
+
+      # Stable identity for the underlying config snapshot, suitable as a cache-key component.
```
timriley:
```suggestion
      # Stable identity for the underlying config snapshot
```

Since this is already called `cache_key`, we probably don't need to repeat ourselves in the description :)

lib/hanami/view/rendering.rb
```diff
@@ -7,28 +7,43 @@ class View
...
+      #
+      # @api private
+      # @since 2.3.0
```
timriley: Let's stop putting `@since` tags on new `@api private` methods. These are for us to use internally only, and putting a `@since` IMO implies some level of stability to external readers of the code that I don't want to promise. (It's private API, after all!)

I need to make a post about this, but in the meantime I'm trying to stop it proliferating, and removing the tags whenever I edit a file.

lib/hanami/view/rendering.rb
```diff
@@ -54,6 +69,10 @@ def part(name, value, as: nil)
       def scope(name = nil, locals) # rubocop:disable Style/OptionalArguments
         scope_builder.(name, locals: locals, rendering: self)
       end
+
+      private
+
+      attr_reader :renderer
```
timriley: I mean, this is probably why I don't like private attr_readers. As a maintainer looking at this class, I see a bunch at the top of the file, and then we have this random one hanging out all the way at the bottom. It's disjointed and it's hard to get a good overview of the class this way.

Maybe we could just leave them at the top, since attr privacy is not really the purpose of this PR?

lib/hanami/view.rb
```diff
@@ -595,16 +590,20 @@ def exposures # rubocop:disable Style/TrivialAccessors
...
 
       locals = locals(rendering, input)
-      output = rendering.template(config.template, rendering.scope(config.scope, locals))
+      output = rendering.template(config_data.template, rendering.scope(scope_class, locals))
 
       if layout
         output = rendering.template(
-          self.class.layout_path(layout),
-          rendering.scope(config.scope, layout_locals(locals))
+          File.join(*[config_data.layouts_dir, layout].compact),
```
timriley: Could we leave the layout_path method back as it was? I think it makes this code here (which is already doing a lot!) easier to follow. One fewer piece of detailed implementation inlined into the method.

lib/hanami/view.rb
```diff
@@ -613,20 +612,26 @@ def call(format: config.default_format, context: config.default_context, layout:
...
+    def rendering(format: config_data.default_format, context: config_data.default_context)
+      Rendering.new(config_data:, format:, context:)
     end
 
     private
 
+    # Frozen Data snapshot of the view's resolved configuration values.
+    # Used for fast hot-path reads during rendering.
+    #
+    # @since 2.3.0
```
timriley:
```suggestion
```

Another since we can drop. This is not just "@api private", it's _fully_ private :)

Gemfile
```diff
@@ -34,6 +36,8 @@ group :benchmarks do
   gem "actionpack"
   gem "actionview"
   gem "benchmark-ips"
+  gem "stackprof", platform: :mri
+  gem "memory_profiler"
```
timriley: Are these not strictly needed for this PR?

lib/hanami/view/rendering.rb
timriley: (Also, for any "since" that is legitimately needed here — now for "@api public" methods only! — it should be `@since 3.0.0`)

... though I think most of your changes are to private API, AFAICT

timriley (commented):
Looking good, @cllns! Left a few thoughts for you, nothing major. Would you mind taking care of the adjustments?

lib/hanami/view/rendering.rb
cllns: Ah yes, agreed! Removed all of them from the files I touched 3eb826b

lib/hanami/view/rendering.rb
cllns: True f9d0394

lib/hanami/view/rendering.rb
cllns: I like doing `private attr_reader :renderer` in my own apps, up top after the public attr_readers. That's not a convention we use though so don't want to add it here. I made this public as you suggested f512095

lib/hanami/view.rb
cllns: Done! 848ecfa

lib/hanami/view.rb
cllns: Done as part of 3eb826b

Gemfile
cllns: Ah yea they were for some temporary benchmarks I wrote in the course of this. Removed in c965462

cllns:
Addressed all feedback! The commit message for 3eb826b renders oddly on GitHub because it thinks `@since` is a username reference, the actual commit message is correct. I'd edit it to use backticks but don't wanna message up the SHAs I posted in replies

timriley (approved):
Thanks @cllns, this is great!

